PR #2, opened 2025-02-24
Revised Vincent contracts

spacesailor24:


packages/vincent-contracts/src/VincentAgentRegistry.sol
DashKash54, line 1: What's the sequence of contract calls to delegate a Role from an App for the first time when the App hasn't been permitted yet?
Is this correct? User PKP calls `permitAppForAgentPkp` then `addRole`? 
So when is the `setAppEnabled()` called? After the User PKP wanna enable the App after disabling it?

packages/vincent-contracts/src/VincentAgentRegistry.sol
```diff
@@ -0,0 +1,542 @@
...
+
+    // ------------------------- App Management -------------------------
+
+    /**
+     * @notice Enable or disable an app
+     * @param agentPkpTokenId The agent PKP token ID
+     * @param appManager The app management wallet address
+     * @param enabled Whether to enable or disable the app
+     */
+    function setAppEnabled(uint256 agentPkpTokenId, address appManager, bool enabled)
```
DashKash54, line 187: The Agent owner also calls this function? `permitAppForAgentPkp`

packages/vincent-contracts/src/VincentAgentRegistry.sol
```diff
@@ -0,0 +1,542 @@
...
+        address appManager,
+        bytes32 roleId,
+        string calldata roleVersion,
+        string[] calldata toolIpfsCids,
+        bytes32[][] calldata policyParams,
+        bytes[][] calldata policyValues
+    ) external onlyPkpOwner(agentPkpTokenId) {
+        require(appManager != address(0), "VincentAgentRegistry: zero address");
+        require(
+            appRegistry.isAppPermittedForAgentPkp(appManager, agentPkpTokenId),
```
DashKash54, line 130: Can we create a abstraction function that permits the Agent in the AppManagement Registry contract if it's being added as a Role? Otherwise the User PKP needs to sign multiple tx and it has to hit the relayer multiple times?

DashKash54 (commented):
Left a few comments

PR #4, opened 2025-02-24
Refactors for demo

Spacesai1or:
Contracts deployed to:

```
VincentAppDelegationRegistry: 0x9B9cfE712a738BC8885a16F63240cCfE17959aF3
VincentAgentRegistry: 0xaE7C442E8d8A6dc07C02A6f41333C2480F28b430
```


DashKash54 (commented):
Left some comments we don't have to make these changes for the Demo though if the features work for now @spacesailor24

spacesailor24:
> Left some comments we don't have to make these changes for the Demo though if the features work for now @spacesailor24

@DashKash54 yea it seems to be working for now. Have tested the entire flow while building out the DCA backend API/service

PR #5, opened 2025-02-25
Heroku & Mongo support

MaximusHaximus:
- Adds configuration to the server for mongo URI
- Defines `engines` in the package.json and a `Procfile` which is required by Heroku
- Updated Mongoose schemas to use Typescript interfaces
- Revised filenames to .ts, module to `esnext`, and module resolution to `bundler` -- this solved some build errors I got from NX when building for deployment to Heroku
- Configured CORS support with the same middleware configuration as the vincent-app-registry backend
- Modified schema for `PurchasedCoin` model to support PopulatedDocument (mongoose functionality) and set appropriate type. Instead of `userId`, it is just `user` -- because it can be resolved using `populate()` to the full user document
- Renamed `start` to `dev` so it's more clear that it uses ts-node transpilation and shouldn't be the entrypoint for the actual running service in production
- Added `timestamps: true` configuration to the Mongoose models so we don't need to manually set `updatedAt` and `createdAt`.
- Use `lean()` to get raw JS objects of documents instead of rich Mongoose model objects when we're returning those objects to the caller


spacesailor24:
Manually incorporated these changes in #6 and merged into `develop`

PR #19, opened 2025-03-14
fix: review changes

anshss:


packages/sdk/src/auth/services/storage-service.ts
```diff
@@ -1,128 +1,74 @@
...
    * @param jwt - The JWT string to store
    */
   async storeJWT(jwt: string): Promise<void> {
-    await this.storage.setItem(Storage.JWT_KEY, jwt);
+    if (!this.isBrowser) {
+      console.warn('Storage operations are not supported in Node environment');
+      return;
+    }
+    try {
+      localStorage.setItem(Storage.JWT_KEY, jwt);
```
MaximusHaximus, line 19: Rather than using a static string for where we store the JWT, we should namespace the key with the app ID.  If multiple apps are hosted on a single domain, each app on that domain must be able to store their own JWT; right now that would be impossible.

packages/sdk/src/auth/services/storage-service.ts
```diff
@@ -1,128 +1,74 @@
...
    */
   async clearJWT(): Promise<void> {
-    await this.storage.removeItem(Storage.JWT_KEY);
+    if (!this.isBrowser) {
+      console.warn('Storage operations are not supported in Node environment');
+      return;
+    }
+    try {
+      localStorage.removeItem(Storage.JWT_KEY);
+    } catch (error) {
```
MaximusHaximus, line 53: :hocho: Just let the error bubble up; logging in catch statements, then re-throwing the error is an anti-pattern that almost always leads to excessive log output.  A good general rule-of-thumb is to only log errors when they are being handled, otherwise every level of your stack will end up logging and re-throwing the _same error_. 

This applies for all of the methods in this class that follow this pattern

packages/sdk/src/auth/services/storage-service.ts
```diff
@@ -1,128 +1,74 @@
...
 
   /**
    * Retrieve the stored JWT token
    * @returns The stored JWT string or null if not found
    */
   async getJWT(): Promise<string | null> {
-    return await this.storage.getItem(Storage.JWT_KEY);
+    if (!this.isBrowser) {
+      console.warn('Storage operations are not supported in Node environment');
+      return null;
```
MaximusHaximus, line 33: This should not just return null. The entire implementation is broken in this case; the error should bubble up to the user.

Rather than repeating console.warn statements in the functions defined here that may not be seen, it would be preferable to throw an error from the class constructor so that the developer is immediately aware that they are trying to run code in an unsupported environment.

MaximusHaximus (changes requested):
1. Looking at the changes we want to make to the `Storage` class (added in-line in comments), I don't think it actually make sense to have that as it's own class at all.  I think it's reasonable, given the hard dependency on local storage, to in-line the calls to localStorage into the Vincent class and just call it directly.  The extra layer of abstraction won't be providing any functionality, but it adds cognitive load and provides a place for bugs to pop up. If we need to provide a storage abstraction in the future, we can add it when we have an actual implementation for 2 storage types that we know we need.
2. I don't see the reference to `window.ethereum` being removed in this diff -- only the global type annotation. `config.ts` appears to still assume that we're using Metamask for getting a signer, even though the delegatee LIT action running is going to run outside of the browser.  Rather than a boolean flag of `isSigner` for the `getContract()` method, we probably just want an argument that _is a signer_ to be passed in.
3. Why do we export all of the 'utils' functions as part of the root SDK API?  Do we actually expect our SDK consumers to call e.g. `splitJWT()`? These should be internal details of how we handle the JWT inside of our SDK, not exported from the package at all.
4. Calling `connect()` on a LitNodeClient is an expensive operation; it looks up the current validators from chain and then handshakes with all of the current nodes -- which can take several seconds.  Both `generateSessionSigs()` and `invokeLitAction()` call it _every time they are called_ -- but a LitNodeClient only needs to connect to the network _once_, when it is first constructed -- from then on you should not call it again.  This is particularly bad because `invokeLitAction()` _calls_ `generateSessionSigs()` -- so executing the LA will take 2x the time it takes to connect to all LIT nodes, _every time it is called!_.
- The consumer of the Vincent SDK should create a LitNodeClient instance and call connect() on it _before using it with the Vincent SDK_.  This would mean that the Vincent SDK methods would not be able to throw all errors that might be encountered while connecting the client. It also gives the consumers of the SDK the ability to _configure the LitNodeClient_ -- e.g. to set `debug` or other settings relating to timeouts to custom values.  Right now, since we construct the client internally, it is impossible for the consumer to configure.

MaximusHaximus:
A follow-up to my review aboe: from our discussions on Friday, I was under the impression there was only one instance of contracts -- `datil` specific -- for Vincent.  But I see `VINCENT_DIAMOND_ADDRESS` is defined with separate diamond addresses labelled `datil-dev-`, `datil-test`, and `datil`.  If we've decided that Vincent is Datil-only, should those other diamonds be removed?

PR #31, opened 2025-03-24
[Example Only] local package chain manager usage in the dashboard

Ansonhkg:
# WHAT

This is an example of using the local chain manager inside the dashboard app


MaximusHaximus:
Closing this as the new contracts SDK is a simpler wrapper approach with a different API

PR #40, opened 2025-03-26
[Example Only] Using Context Provider for contract APIs

Ansonhkg:
# WHAT

An example of setting up a Lit Context provider that initialises the chain manager, so that all children can consume the APIs.

- [LitContext.tsx](https://github.com/LIT-Protocol/Vincent/blob/1d17171cd555f42898765739a7058a7a79ed5392/packages/dashboard/providers/LitContext.tsx) -> Provides a context for the whole app
- StateWrapper.tsx -> Manage states of the fetch/contract calls (eg. isLoading, isEmpty, isError, etc.)
- GetAppsByManager.tsx -> The filename maps 1:1 to the `getAppsByManager`API. It renders a list of apps from the chain manager.

[Here](https://github.com/LIT-Protocol/Vincent/pull/40/files#diff-7023f0455583b0efdb150d2a025917393d6e656a8dd0e0f1ea59f47a5a464995R8-R14) is how your set the return type as return type of the getAppsByManager API.

image

image


Ansonhkg:
Closing as is no longer relevant

PR #46, opened 2025-03-28
V1 Fixes

awisniew207:
Fixed:
- int256 issue when publishing apps
- app creation hydration issue
- allowing policy param updates on a deprecated version
- allowing the option to continue with an existing version
- prepopulating version fields for old and new versions
- prepopulation based on name + datatype, not index
- unified css for the Consent Page
- smarter array form fields on the Consent Page
- allowing null (empty) policy parameter values without defaults
- duplicate ipfs cid permission check
- removal of encoding fallback and default values
- include `appId` in the route
- extended expiration times


awisniew207:
Sorry I wanted to have this as 2 PRs but they got merged :(

PR #48, opened 2025-03-31
Feat/sdk api revision

FedericoAmura:
# Description

This PR cleans the Vincent SDK public interface by:
- hiding unnecessary functions that should not be exported
- cleans some types and function signatures. Removed several unnecessary async calls
- removing storage not used in a backend environment and leaving that concern to consumers where needed (APIs should be stateless and BE apps or FE might want to choose memory, localstorage if available or even DBs)
- avoids constant lit node client re-connections. This could be extremely problematic in APIs serving multiple customers simultaneously


packages/sdk/src/core/vincent.ts
```diff
@@ -1,32 +1,28 @@
...
 export class VincentSDK {
-  private storage: Storage;
-  private consentPageUrl: string;
-  private network: LIT_NETWORKS_KEYS;
+  private readonly consentPageUrl: string;
+  private readonly network: LIT_NETWORKS_KEYS;
 
   constructor(config: VincentSDKConfig = {}) {
-    this.storage = new Storage(config.storage);
     this.consentPageUrl = config.consentPageUrl || 'https://demo.vincent.com';
```
MaximusHaximus, line 24: We should change this default URL should be our heyvincent.ai consent page; this domain actually has something parked on it.

MaximusHaximus (changes requested):
Looking good except for one tweak we should make to the `consentPageUrl` default value :+1:

MaximusHaximus (approved):
Looks good as an initial release 👍

MaximusHaximus (changes requested):
Actually, I was just doing a final pass on the SDK as a whole before merging this PR, and I think we should make some additional changes/fixes:

1. There are some methods on the SDK that point at paths that don't exist:
```
  // Consent Page Management
  openSignInConsentPage(): void {
    const url = new URL('/signin', this.consentPageUrl);
    window.open(url.toString(), '_blank');
  }
  openDelegationControlConsentPage(): void {
    const url = new URL('/delegate', this.consentPageUrl);
    window.open(url.toString(), '_blank');
  }
```
3. Also, I don't think the VincentSDK object should have an `invokeLitAction()` method, given it will _only apply in the context of running tools_, and if it creates a new instance of `DelegateeSigs` every time it's called, then it's also creating a brand new instance of the LitNodeClient every time, because the constructor creates the LitNodeClient instance -- so we're running `connect()` again on every call.

- What do you think about renaming `DelegateeSigs` entirely to `VincentToolRunner`, since that's what it's supposed to be used for -- for backend consumers of the SDK, they would create a single instance of a VincentToolRunner object, and run executeTool on it, so one instance of the LitNodeClient is maintained (edited) 
- What do you think about renaming `VincentSdk` to `VincentAuthManager`, since that is all it is used for once we remove the `invokeLitAction` method?
- `createSignedJWT()` is probably _not useful_ for tool or app developers, since the only source of these signed JWTs currently is the Vincent Dashboard. Should we remove that or make it an internal method?

PR #64, opened 2025-04-07
Adds a `Spending Limit Exceeded` error to Spending Limit Policy

spacesailor24:
See [this thread](https://github.com/LIT-Protocol/Vincent/pull/64#discussion_r2032461146) for context on new error handling

---

IPFS CIDs:

- ERC20_APPROVAL_TOOL_IPFS_ID=`Qma1WPzMy1QLD6TA1hnsHJKtdyxKvKirTd2GMcNewTTWCY`
- UNISWAP_SWAP_TOOL_IPFS_ID=`QmTpsBg8U3uD9KozH1k6dZN3cx4xxiq8gmWmnfWhGVDvmw`
- SPENDING_LIMIT_POLICY_IPFS_ID=`Qme9z19LAnwSJGGGyijprCRfBAaHUuTRbZiiZLFs3JjoY9`


packages/vincent-tools/vincent-tool/src/lib/lit-action-utils/validate-user-tool-policies.ts
```diff
@@ -15,30 +15,50 @@ export const validateUserToolPolicies = async (
...
+                    toolParams,
+                    policy,
+                },
+            });
+        } catch (error) {
+            console.error(`Error executing policy: ${policy.policyIpfsCid} with parameters: ${JSON.stringify(policy.parameters)}`, error);
+
+            return {
+                status: 'error',
+                error: (error as Error).message || String(error)
```
MaximusHaximus, line 55: We should use `JSON.stringify()` here instead of `String()` -- running `String()` on an object will give us `[Object object]` instead of the values of that object (for example)

MaximusHaximus (changes requested):
Just one thing about how we're stringifying error causes, then we can get this merged.

As a follow-up to this, we need to codify how policies work and make it explicit when there is an _error_ vs. when _the policy has simply denied the operation._. We don't want to be throwing runtime errors for cases that we _know_ are just a 'deny this action' state.

Here's what we want moving forward:
https://www.notion.so/litprotocol/Vincent-Policies-API-Spec-1bc6f449b04180b2a241f140b7030b2f

Feedback welcome - we don't need to make these changes to get this shipped and at least use it for testing DCA, but I think our Vincent-Tools helper APIs should conform to this, and we should write our example policies in such a way that the golden path for this is clear to tool/policy devs

packages/vincent-tools/vincent-tool/src/lib/lit-action-utils/validate-user-tool-policies.ts
spacesailor24: [This commit](https://github.com/LIT-Protocol/Vincent/pull/64/commits/59e55d377e28aa1ddfc00f374cd59f8f32e7da29) and [this commit](https://github.com/LIT-Protocol/Vincent/pull/64/commits/e40e98fb845c49961c9847d472ef0568cb80f2cc) address the `String()` issue as well implemented _some_ things from the spec you linked:

- The Policy uses the format:

```ts
{
    allow: bool,
    details: string[]
}
```

Where `details` is populated with all success messages when `allow` is `true`. When `allow` is `false`, `details` is populated with success message up until the policy is violated, then the reason the policy check failed is `push`ed to the end of the array.

- The Lit Action now returns:

```ts
{
    status: 'error' | 'success',
    details: string[]
}
```

Eg:

```ts
{
  "status": "error",
  "details": [
    "Spending limit exceeded",
    "Attempting to spend 158777460000 USD for App ID: 136 when the max daily spending limit is 100000000 USD"
  ]
}
```

or

```ts
{
  "status": "success",
  "details": [
    "Swap transaction hash: 0x1c96416dcf46f95705e27013f692a146af444cb86417b624051be92620e9ca54",
    "Swapped 0.00001 0x4200000000000000000000000000000000000006 for 0x4ed4E862860beD51a9570b96d89aF5E1B0Efefed"
  ]
}
```

spacesailor24: I think using `status: error` for Policy violations isn't ideal, and we probably do something like:

```ts
{
    "status": "policyViolated",
    "policyIpfsCid": "Qm....",
    "violation": "Spending limit exceeded",
    "details": [
    "Attempting to spend 158777460000 USD for App ID: 136 when the max daily spending limit is 100000000 USD"
  ]
}
```

but we should hash this out in the spec, along with some of the other restructuring changes to the Lit Actions to allow for reuse of the util functions outside of the Lit Action

MaximusHaximus (approved):
:+1: We can land this now -- we'll follow-up shortly once we've defined our generic Vincent Tool and Policy handlers

PR #68, opened 2025-04-09
Remove Tool + Policy ENVs

awisniew207:
Allow for an array of policies, remove the ENVs to reduce complexity


packages/dashboard/app/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DCA_POLICY_IPFS_CID = ['QmZrG2DFvVDgo3hZgpUn31TUgrHYfLQA2qEpAo3tnKmzhQ'];
```
MaximusHaximus, line 1: We don't want to sort arrays and compare them for exact matches; there will be different subsets of IPFS cids involved in each consent flow and we don't need them to all match exactly.

I don't think we need to even check for specific _tool_ ipfs CIDs at all -- if we make an index of all _policy_ IPFS CIDs that we know need signing permission, then we should be able to check if any particular policy, independently of any tool they are assigned to, needs signing permission.

I think the logic would be:
- For this consent flow, get a list of all of the IPFS CIDs of all of the **_policies_** that the current app version is requesting consent for
- Map across them and check them against our map of `IPFS_POLICIES_THAT_NEED_SIGNING` -- and if there is an entry in the map matching the policy IPFS CID, check they have permission and add it if they don't.

Something like this:
```typescript
const IPFS_POLICIES_THAT_NEED_SIGNING = {
  '<ipfsCid1>': { 'description': 'Spending policy published 1/1/2025, commit hash XYZ' },
  '<ipfsCid2>': { 'description': 'Spending policy published 1/12/2025, commit hash ABC' },
};

for(ipfsCid in currentAppVersionPolicies) {
  if(ipfsCid in IPFS_POLICIES_THAT_NEED_SIGNING) {
    console.log(`Checking for existing sign permission for policy ${ipfsCid}, ${IPFS_POLICIES_THAT_NEED_SIGNING[ipfsCid].description}`);
  }
}
```

Note that we want to use an object, where the ipfs cids are keys in the map, rather than an array -- because searching sequentially through arrays is slow whereas looking up keys on an object is very fast.  As a general best practice, if the _order_ of data is not important, and we don't need to iterate across it in _specific order_ don't keep it in arrays :)

MaximusHaximus (changes requested):
Added a comment in-line -- I think we need this logic to work a bit differently than we currently do, and I'm not sure we even need the tool ipfs CIDs at all, since the _policies_ are the ones that are special and we should be able to know the policy IPFS cids in the context where we are processing these.

packages/dashboard/app/constants.ts
awisniew207: I think this version is a lot better, and the changes weren't hard to make with the existing code. I've updated the logic for this to work as described, and I'll attach a screenshot of it working during testing :smile_cat: 

image

packages/dashboard/components/consent/utils/consentTransactionUtils.ts
```diff
@@ -148,29 +150,36 @@ export const addPermittedActions = async (
...
+          console.log(`Adding sign permission for policy ${ipfsCid}, ${IPFS_POLICIES_THAT_NEED_SIGNING[ipfsCid].description}`);
+          const tx = await litContracts.addPermittedAction({
+            ipfsId: ipfsCid,
             pkpTokenId: agentPKPTokenId,
             authMethodScopes: [AUTH_METHOD_SCOPE.SignAnything],
           });
+          console.log(`Added sign permission for policy ${ipfsCid} - Transaction hash: ${tx.transactionHash}`);
         }
+      } catch (error) {
+        console.error(`Error adding DCA policy permission for ${ipfsCid}:`, error);
```
MaximusHaximus, line 171: We can't just ignore failures in this flow to check for or add new tool policy permissions -- if we do this, the user will appear to sign up successfully, we'll provide a JWT to the app via redirect, _but when the app tries to actually execute tools for the user, they will fail due to invalid delegations_.

Consider that, in the scenario I just described, _the only resolution the app can provide to the user_ would be for the user to return _back to this screen_ and go through _this process again_ to fix the problem. Then, consider that _it might fail again_, but if so, this code will once again ignore the failure, and the user will end up 'signed in' to the app, but their ability to actually use the app will still be broken.

We need to track an error state of some kind, and make the consent page render that error state if it is set -- that wya, If this logic throws an error,  the consent page can tell the user there was an error and give them a 'retry' button or something so they can re-run this code from the top-down, rather than us just ignoring the error and sending them off to an app that we can reasonably assume (right here in this code) will be in a broken state.

packages/dashboard/components/consent/utils/consentTransactionUtils.ts
```diff
@@ -193,6 +202,8 @@ export const addPermittedActions = async (
         pkpTokenId: agentPKPTokenId,
         authMethodScopes: [AUTH_METHOD_SCOPE.SignAnything],
       });
+      console.log(`Added permission for ${ipfsCid} - Transaction hash: ${tx.transactionHash}`);
+
     } catch (error) {
       console.error(
         `Error adding permitted action for IPFS CID ${ipfsCid}:`,
```
MaximusHaximus, line 209: See my comment above -- if this code fails, we must set an error state of some kind and reflect the failure to the user, otherwise we are just sending the user off to an app that isn't actually functional, and where the app developer has no way to actually resolve the issue other than sending the user right back here.

MaximusHaximus (changes requested):
:+1: This is looking good from the policy iteration/checks perspective -- but we should eliminate the silent swallowing of errors in this logic to ensure we actually know when things go wrong at the point that they do go wrong and can still be resolved before the user ends up redirect to an app that isn't functional -- see comments below

PR #71, opened 2025-04-10
Consent Page Warnings for Disabled Versions

awisniew207:
Adding warnings:

image
image

Styles may change soon if you think these are ugly, but the warning is needed.

### DIFF
--- a/packages/dashboard/components/consent/components/authForm/ParameterUpdateModal.tsx
+++ b/packages/dashboard/components/consent/components/authForm/ParameterUpdateModal.tsx
@@ -1,14 +1,28 @@
 import React from 'react';
+import { useVersionEnabledCheck } from '../../hooks/useVersionEnabledCheck';
 
 interface ParameterUpdateModalProps {
   isOpen: boolean;
   onContinue: () => void;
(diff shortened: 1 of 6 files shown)


packages/dashboard/components/consent/components/authForm/ParameterUpdateModal.tsx
```diff
@@ -1,13 +1,27 @@
...
+  onContinue, 
+  onUpdate, 
+  appName, 
+  permittedVersion 
+}: ParameterUpdateModalProps) => {
+  const { isVersionEnabled } = useVersionEnabledCheck({
+    versionNumber: permittedVersion
+  });
+  
+  const isDisabled = isVersionEnabled === false;
```
FedericoAmura, line 23: nit really:

```suggestion
  const isDisabled = !isVersionEnabled;
```
Or better just use it directly with the `!` below to avoid making a new variable

awisniew207: Fixed!

FedericoAmura (approved):

PR #72, opened 2025-04-10
Notification + Error updates

awisniew207:
As the title says, this is a simple fix so that users can't permit a disabled latest version of an application. If they do, the gas estimation will fail since the contract doesn't permit this, so we stop them before they get that far.

Also including better Stytch notifications in this PR, since these are both user notifications


packages/dashboard/components/consent/components/AuthenticatedConsentForm.tsx
```diff
@@ -256,6 +258,27 @@ export default function AuthenticatedConsentForm({
...
+  // Add debugging to track version information
+  useEffect(() => {
+    if (versionInfo) {
+      console.log('Version info structure in AuthenticatedConsentForm:', {
+        hasVersionInfo: !!versionInfo,
+        appVersion: versionInfo.appVersion,
+        versionHex: versionInfo.appVersion.version.hex,
+        parsedVersion: versionInfo.appVersion.version.hex ? parseInt(versionInfo.appVersion.version.hex, 16).toString() : 'Unknown',
+        appInfoLatestVersion: appInfo?.latestVersion
+      });
```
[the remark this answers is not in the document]
awisniew207: Sorry this slipped by, since deleted

packages/dashboard/components/consent/components/StytchOTP.tsx
```diff
@@ -67,8 +98,19 @@ const StytchOTP = ({ method, authWithStytch, setView }: StytchOTPProps) => {
       }
       
       await authWithStytch(response.session_jwt, response.user_id, method);
-    } catch (err) {
+    } catch (err: any) {
       console.error(`Error authenticating with ${method} OTP:`, err);
+      let errorMessage = 'Failed to verify code. Please try again.';
+      
+      if (err instanceof z.ZodError && err.errors.length > 0) {
+        errorMessage = err.errors[0].message;
```
spacesailor24, line 107: Is `err.errors.length > 1`? In which case we'd only show a single error?

awisniew207: Yeah this will show the first error, but it's specific to the zod validation schema failing. If it does fail, we provide the error from the schema (which is always only one): 

image

spacesailor24 (approved):

PR #73, opened 2025-04-11
feat: add authorized app and user info to jwt

FedericoAmura:


packages/sdk/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2018",
-    "module": "nodenext",
+    "module": "esnext",
```
FedericoAmura, line 4: DCA FE was throwing `the requested module does provide an export named ...` when importing from `@lit-protocol/vincent-sdk` because of this

This config should maximize compatibility as "nodenext" is focused on nodejs and relies on files extensions and package.json type field

MaximusHaximus: Unfortunately, using `esnext` breaks the SDK entirely for NodeJS users, because it emits raw ESM modules, (no TSLib helpers), and when NodeJS encounters imports from the SDK with raw ESM syntax, it considers them invalid for a few reasons -- the primary problems are that it doesn't allow directory imports, and it requires that _all files have file extensions_.  The last part is particularly problematic because it requires they have _.js file extensions if the files are in `.dist/*.js` -- but obviously we can't change all of our imports in the SDK to import `.js` files while we are in fact writing TS code.  This was actually a necessary and intentional recent change -- https://github.com/LIT-Protocol/Vincent/pull/70 -- because running the SDK in our Heroku service for DCA was broken with the existing `esnext` configuration.

See this post for more details about this that are specifically for library authors: https://blog.andrewbran.ch/is-nodenext-right-for-libraries-that-dont-target-node-js/ -- from the [author of typescriptlang's "Choosing Compiler Options"](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library).  In particular, note that using `Bundler` in _our_ tsconfig is parasitic -- it forces consumers of our SDK to use a bundler for compatibility with our SDK.

To summarize from the article:
> I think I would summarize as:
> 
> nodenext is the right option for authoring libraries, because it prevents you from emitting ESM with module specifiers that only work in bundlers but will crash in Node.js. When writing conventional code, using common sense, and relying on high-quality dependencies, its output is usually highly compatible with bundlers and other runtimes.
> There is no TypeScript option specifically meant to enforce patterns that result in maximum cross-environment portability.

I think the issue we're experiencing with locally linking is a vite-specific issue related to `pnpm` using symbolic links to import the package -- I tested publishing up the SDK _without_ the changes you made as `0.0.6-beta.1`, and local development worked as expected, even though the filesystem shape is identical in the published package as the locally linked one. I agree that we should figure out the local dev issue, but no matter what combination of tools, _i can't make these setting work on NodeJS without running my code through a bundler_.  Since the SDK includes code that is specifically designed for backend use (running Vincent tools) and the issue we're seeing with local dev is specific to _linking the deps w/ vite_, I think we should go with the wider compatibility configuration and spend some time figuring out how to get Vite to pay nicer with pnpm local linking (or use an alternative method for local dev that doesn't involve pnpm's symlinks maybe?) rather than force all SDK consumers to use a bundler for their Node code.

packages/sdk/src/express-authentication-middleware/express.ts
```diff
@@ -13,18 +13,25 @@ function assertAuthenticatedRequest(req: Request): asserts req is AuthenticatedR
...
+  function isDefinedObject(value: unknown): value is object {
+    return typeof value === 'object' && value !== null;
+  }
 
   if (
     typeof rawJWT !== 'string' ||
-    typeof pkpAddress !== 'string' ||
-    typeof decodedJWT !== 'object' ||
-    decodedJWT === null
+    !isDefinedObject(app) ||
```
MaximusHaximus, line 31: 🤔 `pkpAddress` felt like a convenient minor duplication -- but reading this, it feels like it's getting out of hand -- maybe we should encourage devs to just reference the decoded JWT instead of copying all the data from the payload directly on to the user -- WDYT? Then they would have a consistent reference to the VincentJWT type everywhere they need data from it...

packages/sdk/src/jwt/core/validate.ts
```diff
@@ -63,7 +63,7 @@ export function verifyJWT(jwt: string, expectedAudience: string): VincentJWT {
     const s = signatureBytes.slice(32, 64);
 
     // Process public key
-    let publicKey = decoded.payload.pkpPublicKey;
+    let publicKey = decoded.payload.pkp.publicKey;
```
MaximusHaximus, line 66: Since this is validating what we just deserialized from a string, we should check that all of the expected Vincent JWT payload properties exist on the JWT and are well formed (like we were doing for `pkpPublicKey` before this change) and not us `as VincentJWT` when we're not actually sure at runtime that it is one.

MaximusHaximus (changes requested):
Added a few things in-line -- see especially remarks re: esnext vs. nodenext

packages/sdk/src/jwt/types.ts
```diff
@@ -20,19 +20,35 @@ export interface JWTConfig {
...
+ * @property {string} app - The app associated with the JWT.
+ * @property {string} pkp - The PKP associated with the JWT.
+ * @property {string} authentication - The authentication method used to generate the JWT.
  */
 export interface VincentJWTPayload extends JWTPayload {
-  pkpPublicKey: string;
-  pkpAddress: string;
+  pkp: IRelayPKP;
+  app: {
+    id: string;
```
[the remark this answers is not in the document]
FedericoAmura: No. In JS land they come from the url param in [this hook](https://github.com/LIT-Protocol/Vincent/blob/main/packages/dashboard/components/consent/hooks/useUrlAppId.ts#L15)
And in solidity land, they are a uint256. So they cannot fit into a JS Number

packages/sdk/tsconfig.json
FedericoAmura: Fixed by returning to the previous build config and adding an optimization step in the DCA FE which solves this

packages/sdk/tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noImplicitThis": true,
+    "moduleResolution": "bundler",
```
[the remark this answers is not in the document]
FedericoAmura: Gone

packages/sdk/src/jwt/core/validate.ts
FedericoAmura: Consumers now would have to use the decoded payload directly. That way we will avoid duplicating references

packages/sdk/src/jwt/types.ts
```diff
@@ -20,19 +20,35 @@ export interface JWTConfig {
...
+    version: number;
+  };
+  authentication: {
+    type: string;
+    value?: string;
```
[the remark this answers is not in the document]
MaximusHaximus: IIRC we added a prompt to WebAuthN that tracks a user-provided 'username' or similar label -- we should get that in the auth info.  cc: @awisniew207 can we get a user displayable value for people using WebAuthN login?

packages/sdk/src/express-authentication-middleware/express.ts
```diff
@@ -1,6 +1,7 @@
 import type { NextFunction, Request, Response } from 'express';
 
-import { jwt } from '@lit-protocol/vincent-sdk';
```
MaximusHaximus, line 3: 💖 Good catch, I just spotted this tonight and was going to open a PR to fix that reference 🍭

packages/sdk/src/jwt/core/validate.ts
```diff
@@ -62,8 +68,13 @@ export function verifyJWT(jwt: string, expectedAudience: string): VincentJWT {
     const r = signatureBytes.slice(0, 32);
     const s = signatureBytes.slice(32, 64);
 
+    const { app, authentication, pkp } = decoded.payload;
+    if (!isDefinedObject(app) || !isDefinedObject(authentication) || !isDefinedObject(pkp)) {
+      throw new Error(`Missing required fields in JWT payload`);
+    }
```
MaximusHaximus, line 74: 1. For consistency with the rest of our error syntax, we should still prefix this w/ the appropriate `JWT_ERROR` string.
2. Also, for the benefit of someone debugging any issue and for us to identify what's wrong if we get a report of us providing invalid JWTs to consumers, it would be good to be verbose about what was missing that we expected to be there.

```suggestion
    const { app, authentication, pkp } = decoded.payload;

    if (!isDefinedObject(app)) {
      throw new Error(`${JWT_ERROR.INVALID_JWT}: Missing "app" field in JWT payload.`);
    }

    if (!isDefinedObject(authentication)) {
      throw new Error(`${JWT_ERROR.INVALID_JWT}: Missing "authentication" field in JWT payload.`);
    }

    if (!isDefinedObject(pkp)) {
      throw new Error(`${JWT_ERROR.INVALID_JWT}: Missing "pkp" field in JWT payload.`);
    }
```

3. This check should happen outside of this `try{}` block, above this where we do the other verifications, so that we don't convert these errors into `${JWT_ERROR.INVALID_SIGNATURE}: Invalid signature: ${(error as Error).message}` on L103 below even though missing properties on the JWT isn't a bad signature issue.

packages/sdk/src/jwt/core/validate.ts
```diff
@@ -100,21 +111,5 @@ export function verifyJWT(jwt: string, expectedAudience: string): VincentJWT {
...
-
-  return {
-    ...decoded,
-    payload: {
-      ...decoded.payload,
-      pkpPublicKey,
-      pkpAddress: ethers.utils.computeAddress(pkpPublicKey),
-    },
-  };
+  return didJWT.decodeJWT(jwt) as VincentJWT;
```
MaximusHaximus, line 114: This used to be type-safe since we checked the payload and threw an error if it was malformed -- but now, we are asserting the JWT is a VincentJWT here when it might actually not include any Vincent specific properties.

We should probably _move the validation that checks to be sure the `payload` includes Vincent-specific properties into this function_ so that we don't need to use `as VincentJWT` here -- TS should know it's a VincentJWT as a result of our validations. Since we call `decodeJWT` inside of `verify`, that will resolve the other issue we have with these checks happening inside of the signature verification in that function also.

To re-cap
1. `decodeJWT()` should return a token that is _well formed_, but _might be expired_. If the JWT is not well-formed, it should `throw`.
2. `verify()` should `decodeJWT()`, which will `throw` if the token is _malformed_.  However, it _also_ throws if the signature is incorrect _or_ the token is expired.

MaximusHaximus (changes requested):
Just a couple things in-line for consistency with our existing approach.

packages/sdk/src/jwt/core/validate.ts
FedericoAmura: Good points

packages/sdk/src/jwt/core/validate.ts
FedericoAmura: Better point than the one above

packages/sdk/src/jwt/types.ts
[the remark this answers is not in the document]
FedericoAmura: You are correct, the browser asks for a name when saving the credential

But the js sdk asks for a credential using the domain, not its name (when using stytch the dashboard is saving the requested email/phone) so any credential for that domain could be used really

Also, when getting the credential name (at the js sdk) we only get a hash of it, and the js sdk [encodes it just in case](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts#L157) before putting it **inside** the access token

Screenshot from 2025-04-16 15-52-32

I think this is doable but very costly and without more benefit other than consistency (always having a `value`) if we don't find a use case for that encoded hash

MaximusHaximus (approved):
:+1: Let's do it! Thanks for the iteration, I like where we landed on this API-wise :+1:

PR #76, opened 2025-04-15
Fix display error, remove SessionValidator

awisniew207:
- Cleaned up layouts
- Removed SessionValidator, unified sign in/log in logic
- Cleaner index


packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -23,13 +25,58 @@ const CenteredContainer = ({ children }: { children: React.ReactNode }) => (
...
+    isProcessing,
+    error: readError
+  } = useReadAuthInfo();
+  
+  // State to show existing account option
+  const [showExistingAccount, setShowExistingAccount] = useState(false);
+  
+  // Check for existing auth info once the validation process is complete
+  useEffect(() => {
+    if (isProcessing) return;
```
MaximusHaximus, line 57: Show a spinner or loading text or something?

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -23,13 +25,58 @@ const CenteredContainer = ({ children }: { children: React.ReactNode }) => (
...
+    if (validatedSessionSigs) {
+      setShowExistingAccount(true);
+    }
+  }, [validatedSessionSigs, isProcessing]);
+  
+  // Handle using existing account
+  const handleUseExistingAccount = () => {
+    if (!validatedSessionSigs || !authInfo) return;
```
MaximusHaximus, line 65: If either of these are not set, how would the user even see a button to perform this action?

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -23,13 +25,58 @@ const CenteredContainer = ({ children }: { children: React.ReactNode }) => (
...
+    
+    // Use the existing session and auth info
+    setSessionSigs(validatedSessionSigs);
+    setuserPKP(authInfo.userPKP);
+    setAgentPKP(authInfo.agentPKP);
+    setShowExistingAccount(false);
+  };
```
MaximusHaximus, line 72: 🔪 🔪 there's no need to set the same data into multiple duplicate variable/state holders here. All we need to do is set the flag for `showExistingAccount` -- kill all 3 of these setters and reference the values directly from the object exposed from the `useReadAuthInfo()` hook

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -46,26 +95,26 @@ export default function IndexView() {
...
+    if (newPKP) {
+      setuserPKP(newPKP);
     }
-  }, []);
-
-  // Function to generate session signatures on-demand
+  }
+  
+  // Generate session signatures on-demand
   const generateSessionSigs = useCallback(async () => {
```
MaximusHaximus, line 118: We're generating session sigs inside the `useReadAuthInfo()` hook now -- all of this should be :hocho:?

packages/dashboard/components/consent/views/SignUpView.tsx
```diff
@@ -0,0 +1,148 @@
...
+          </svg>
+          <a href="https://litprotocol.com" target="_blank" rel="noopener noreferrer" className="flex items-center">
+            Protected by <Image src="/wordmark.svg" alt="Lit" width={15} height={9} className="ml-1" />
+          </a>
+        </p>
+      </div>
+    </div>
+  );
+
+  switch (authMethodType) {
```
MaximusHaximus, line 135: 💖

packages/dashboard/components/consent/utils/validateSession.ts
```diff
@@ -0,0 +1,77 @@
...
+      const litWalletSig = localStorage.getItem('lit-wallet-sig');
+      if (!litWalletSig) {
+        return;
+      }
+
+      // Create lit resources for action execution and PKP signing
+      const litResources = [
+        new LitActionResource('*'),
+        new LitPKPResource('*'),
+      ];
```
MaximusHaximus, line 20: :hocho: Inline this -- it's only referenced in one place, this is redundant

packages/dashboard/components/consent/utils/validateSession.ts
```diff
@@ -0,0 +1,77 @@
...
+        expiration: new Date(Date.now() + 1000 * 60 * 60 * 24).toISOString(),
+        sessionKey,
+        sessionKeyUri: `lit:session:${sessionKey.publicKey}`,
+        sessionCapabilityObject,
+        nonce: Date.now().toString(),
+      });
+
+      if (walletSig) {
+        await litNodeClient.connect();
+        const attemptedSessionSigs = await litNodeClient.getSessionSigs({
```
MaximusHaximus, line 41: `getSessionSigs()` always returns a `SessionSigsMap` or throws an error -- 🔪  `attemptedSessionSigs`

packages/dashboard/components/consent/utils/validateSession.ts
```diff
@@ -0,0 +1,77 @@
+import { LitPKPResource } from "@lit-protocol/auth-helpers";
```
MaximusHaximus, line 1: Let's call this `getValidSessionSigs()` (and adjust the filename accordingly) for clarity

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -165,186 +216,76 @@ export default function IndexView() {
...
+    // If authenticated with a new PKP and session sigs
+    if (userPKP && sessionSigs) {
+      // Save the PKP info in localStorage for SessionValidator to use
+      try {
+        updateAuthInfo({
+          agentPKP,
+          userPKP
+        });
+      } catch (error) {
+        console.error('Error saving PKP info to localStorage:', error);
```
MaximusHaximus, line 252: Is it _really_ not a critical error if calling `updateAuthInfo()` throws a hard runtime error? This is swallowing the failure entirely, but it feels like if this fails, their authInfo is in a totally inconsistent state and they may need to re-log in, right?

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -23,13 +25,58 @@ const CenteredContainer = ({ children }: { children: React.ReactNode }) => (
...
+    setSessionSigs(validatedSessionSigs);
+    setuserPKP(authInfo.userPKP);
+    setAgentPKP(authInfo.agentPKP);
+    setShowExistingAccount(false);
+  };
+  
+  // ------ NEW AUTHENTICATION FLOW ------
+  
+  // Authentication state
+  const [sessionLoading, setSessionLoading] = useState<boolean>(false);
```
MaximusHaximus, line 77: How is this different than `authLoading` just under this? I think :hocho: we only need the state exposed from our new hook, and the loading from `useAuthenticate` now, yes?  

Also, `useAccounts()` can only work if someone is logged in with a valid authInfo, right? Should that even exist in a context where authInfo might not be set at all?

packages/dashboard/components/consent/pages/index.tsx
```diff
@@ -154,8 +197,16 @@ export default function IndexView() {
       
       return () => clearTimeout(timeout);
     }
-  }, [authLoading, accountsLoading, sessionLoading, loadingMessage]);
+  }, [authLoading, accountsLoading, sessionLoading, isProcessing, loadingMessage]);
+  
+  // ------ CLEANUP ------
 
+  const handleSignOut = async () => {
+    cleanupSession();
```
MaximusHaximus, line 205: `cleanupSession()` appears to be a duplicate of `clearAuthInfo()` -- we should remove `cleanupSession` and just us `clearAuthInfo()` anywhere `cleanupSession` was previously being referenced

packages/dashboard/components/consent/hooks/useAuthInfo.ts
```diff
@@ -11,29 +11,50 @@ export interface AuthInfo {
...
+          setAuthInfo(parsedAuthInfo);
+        }
+        const sigs = await validateSession();
+        if (sigs) {
+          setSessionSigs(sigs);
+        }
+      } catch (error) {
+        console.error('Error retrieving auth info:', error);
+        setError(error as string);
+        localStorage.removeItem(AUTH_INFO_KEY);
```
MaximusHaximus, line 49: This should be using `clearAuthInfo()` for consistency

packages/dashboard/components/consent/hooks/useAuthInfo.ts
```diff
@@ -11,29 +11,50 @@ export interface AuthInfo {
...
+    const loadAuthInfo = async () => {
+      try {
+        const storedAuthInfo = localStorage.getItem(AUTH_INFO_KEY);
+        if (storedAuthInfo) {
+          const parsedAuthInfo = JSON.parse(storedAuthInfo) as AuthInfo;
+          setAuthInfo(parsedAuthInfo);
+        }
+        const sigs = await validateSession();
+        if (sigs) {
+          setSessionSigs(sigs);
```
MaximusHaximus, line 44: The code in the `useAuthInfo()` hook doesn't remove the `lit-wallet-sig` key from localstorage -- it _only removes the authInfo_.  Yet, it could be that `validateSession` returned `null` because `isValid` was `false`.  I think if we want to force the user to log in if `validateSession` encountered an expired walletSig, we need to clear authInfo and wallet sig here if `validateSession` returned null because of an expired wallet sig, right?

MaximusHaximus (changes requested):
Added comments in-line.  I also noticed that `updateAuthInfo` swallows errors encountered trying to read existing auth info and write it back to localStorage, which could leave someone in an inconsistent state (e.g. their walletSig might be for a different authInfo or something). We should remove any silent squashing of errors in code paths where someone would be left in an inconsistent state and ensure that the error bubbles up to the consumer -- swallowing failures to track critical data like authentication info is not good..

packages/dashboard/components/consent/hooks/useAuthInfo.ts
awisniew207: Will throw when building:
```
./components/consent/hooks/useAuthInfo.ts
56:9  Error: React Hook "useClearAuthInfo" is called in function "loadAuthInfo" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks
```

packages/dashboard/components/consent/utils/validateSession.ts
awisniew207: Changed to:

```
    const sessionCapabilityObject =
      await litNodeClient.generateSessionCapabilityObjectWithWildcards(
        [
          new LitActionResource('*'),
          new LitPKPResource('*'),
        ]
      );
```

packages/dashboard/components/consent/pages/index.tsx
awisniew207: Not necessarily, this is for the case where the user is generating the session signatures using an `AuthMethod`, not a `lit-wallet-sig`

packages/dashboard/components/consent/pages/index.tsx
awisniew207: Good point, changed to:
```
  const handleUseExistingAccount = () => {
    setShowExistingAccount(false);
  };
```

packages/dashboard/components/consent/utils/validateSession.ts
awisniew207: Could we have some constant here to store them? We later perform these using the session signatures, which might look pretty ugly without a constant:
```
      const validationResult = await validateSessionSigs(
        attemptedSessionSigs
      );

      if (validationResult.isValid) {
        return attemptedSessionSigs;
      } else {
        return null;
      }
```

I do agree to remove this though:
```
      if (!attemptedSessionSigs) {
        return;
      }
```

packages/dashboard/components/consent/pages/index.tsx
awisniew207: When `isProcessing` is `true`:
```
  const renderContent = () => {
    // Handle loading states first
    if (authLoading || accountsLoading || sessionLoading || isProcessing) {
      return (
        <Loading 
          copy={loadingMessage} 
          isTransitioning={isTransitioning}
        />
      );
    }
 ```
 So we briefly show:
 
image

File: packages/dashboard/components/consent/hooks/useAuthInfo.ts (hunk 13 above)
Comment (awisniew207) [reply packages/dashboard/components/consent/hooks/useAuthInfo.ts:44]: This is what `cleanupSession()` was doing, calling `disconnectWeb3()` which removes the `lit-wallet-sig`. I've just added `disconnectWeb3()` to the `UseClearAuthInfo` after it removes the `lit-auth-info`

File: packages/dashboard/components/consent/pages/index.tsx (hunk 9 above)
Comment (awisniew207) [reply packages/dashboard/components/consent/pages/index.tsx:252]: Changed to have `updateAuthInfo` include an error which will be returned, currently shown in a popup (for now):
```
catch (error) {
        console.error('Error saving PKP info to localStorage:', error);
        showError(error as Error, 'Authentication Error');
        return (
          <LoginMethods
            authWithEthWallet={authWithEthWallet}
            authWithWebAuthn={authWithWebAuthn}
            authWithStytch={authWithStytch}
            registerWithWebAuthn={handleRegisterWithWebAuthn}
          />
        );
      }
```

awisniew207: I guess we can also remove the `lit-auth-info` if this fails?

packages/dashboard/components/consent/pages/index.tsx
awisniew207: `authLoading` manages the state exposed from the `useAuthenticate` hook, while the `sessionLoading` manages the state of the session signature generation which uses the `AuthMethod` as a result of the `useAuthenticate` hook. `useAccounts` is also based on the premise that there is a valid `Authmethod`, so it isn't dependent on `lit-auth-info`

MaximusHaximus (approved):
:+1: Let's get this landed!

spacesailor24 (approved):
Have just one unresolved bug, but creating a separate ticket for it so it can be fixed in a subsequent PR

[Linear ticket](https://linear.app/litprotocol/issue/LIT-4369/zod-schema-validation-when-readwriting-to-local-storage)

PR #77, opened 2025-04-16
init(dashboard): withdraw page

awisniew207:
WIP, details soon


packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  if (!ethers.utils.isAddress(withdrawAddress)) {
+    showStatus('Invalid withdrawal address', 'error');
+    return;
+  }
+
+  if (selectedToken) {
+    if (selectedToken.symbol === 'ETH') {
+      const parsedAmount = ethers.utils.parseUnits(withdrawAmount, selectedToken.decimals);
+      if (parsedAmount.eq(selectedToken.rawBalance)) {
```
FedericoAmura, line 150: Let's apply this check also if it is greater, same as a few lines below

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,236 @@
...
+    await litNodeClient.connect();
+
+    const pkpWallet = new PKPEthersWallet({
+      pkpPubKey: agentPKP.publicKey,
+      litNodeClient: litNodeClient,
+      controllerSessionSigs: sessionSigs,
+      rpc: BASE_MAINNET_RPC
+    })
+
+    await pkpWallet.init();
```
FedericoAmura, line 179: Can we set this in the parent? Setting up a `LitNodeClient` is a very expensive operation and makes the user have to wait for a looooong time. Likely this can be prepared upfront

packages/dashboard/app/withdraw/utils/hiddenTokens.tsx
```diff
@@ -0,0 +1,117 @@
...
+            return;
+        }
+
+        const newHiddenTokens = [...hiddenTokens, lowercaseAddress];
+        setHiddenTokens(newHiddenTokens);
+
+        try {
+            localStorage.setItem(HIDDEN_TOKENS_KEY, JSON.stringify(newHiddenTokens));
+        } catch (error) {
+            console.error('Error saving to localStorage', error);
```
FedericoAmura, line 91: Don't swallow and hide the error from the user

packages/dashboard/app/withdraw/utils/hiddenTokens.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    const hideToken = useCallback((address: string) => {
+        if (address === 'ETH') return; // Don't allow hiding ETH
+
+        const lowercaseAddress = address.toLowerCase();
+        try {
+            ethereumAddressSchema.parse(lowercaseAddress);
+        } catch (validationError) {
+            console.error('Invalid address format:', validationError);
+            return;
```
FedericoAmura, line 82: Don't swallow and hide the error from the user

packages/dashboard/app/withdraw/utils/hiddenTokens.tsx
```diff
@@ -0,0 +1,117 @@
...
+        
+        const newHiddenTokens = hiddenTokens.filter(
+            (tokenAddr: string) => tokenAddr !== lowercaseAddress
+        );
+        setHiddenTokens(newHiddenTokens);
+
+        try {
+            localStorage.setItem(HIDDEN_TOKENS_KEY, JSON.stringify(newHiddenTokens));
+        } catch (error) {
+            console.error('Error saving to localStorage', error);
```
FedericoAmura, line 106: Don't swallow and hide the error from the user

packages/dashboard/app/withdraw/utils/layout.tsx
```diff
@@ -0,0 +1,37 @@
...
+let stytchClient: any = null;
+if (typeof window !== 'undefined') {
+  try {
+    const publicToken = process.env.NEXT_PUBLIC_STYTCH_PUBLIC_TOKEN;
+    if (publicToken) {
+      stytchClient = createStytchUIClient(publicToken);
+    }
+  } catch (error) {
+    console.error('Error initializing Stytch client:', error);
+  }
```
FedericoAmura, line 18: If this fails we definitely don't want to hide the error from the user. It is a great candidate to crash the app as it is basically 66% of our auth methods

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,187 @@
...
+    let gasLimit;
+    try {
+      const estimateTransaction = {
+        from: await pkpWallet.getAddress(),
+        to: recipientAddress,
+        value: amount
+      };
+      
+      // 10% buffer
+      const estimatedGas = await provider.estimateGas(estimateTransaction);
```
FedericoAmura, line 43: Needed gas for an ETH transfer is always 21_000 (with a 10% margin = 23_100 as we saw in basescan). No need to estimate here

packages/dashboard/app/withdraw/utils/alchemyUtils.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const tokenBalancesList = tokensData.result?.tokenBalances || [];
+    const nonZeroBalances = tokenBalancesList.filter((token: AlchemyTokenBalance) => {
+      if (token.error) return false;
+      const balanceBN = ethers.BigNumber.from(token.tokenBalance);
+      return !balanceBN.isZero();
+    });
+
+    // If we have any non-zero balances, we need to get the token metadata
+    if (nonZeroBalances.length > 0) {
+      const tokenDetailsPromises = nonZeroBalances.map(async (token: AlchemyTokenBalance) => {
```
FedericoAmura, line 96: Moralis gives us all the data in one single call. Definitely worth using their API as this can get very annoying with high amounts of tokens

https://docs.moralis.com/web3-data-api/evm/reference/wallet-api/get-token-balances-by-wallet?address=0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045&chain=base&token_addresses=[]

packages/dashboard/app/withdraw/utils/alchemyUtils.tsx
```diff
@@ -0,0 +1,171 @@
...
+const formatBalance = (balance: ethers.BigNumber, decimals: number): string => {
+  return ethers.utils.formatUnits(balance, decimals);
+};
+
+export const fetchTokenBalances = async (ethAddress: string): Promise<TokenBalanceResult> => {
+  try {
+    const provider = new ethers.providers.JsonRpcProvider(BASE_RPC_URL);
+
+    // Check ETH balance without using Alchemy
+    const ethBalance = await provider.getBalance(ethAddress);
```
FedericoAmura, line 45: Fetching eth balance should be another function. It does not come from alchemy and by being a different entity can be requested in parallel

packages/dashboard/app/withdraw/utils/layout.tsx
awisniew207: I can refactor a more comprehensive fix for this in another PR, but it'll require some more time. This issue should only happen when Stytch's service is down since our ENV for it will (should) never change

packages/dashboard/app/withdraw/utils/transactionService.ts
awisniew207: We'll have to consider the L1 fee as well, so that being the case I've made a new helper function:

```
export async function calculateEthGasCosts(): Promise<{
  totalCost: ethers.BigNumber;
  gasLimit: ethers.BigNumber;
  gasPrice: ethers.BigNumber;
}> {
  const baseProvider = new ethers.providers.JsonRpcProvider(BASE_MAINNET_RPC);
  const ethProvider = new ethers.providers.JsonRpcProvider(ETH_MAINNET_RPC);

  // Get current gas prices for L1 and L2
  const l1GasPrice = await ethProvider.getGasPrice();
  const l2GasPrice = await baseProvider.getGasPrice();
  
  // Fixed gas limits
  const l1GasLimit = ethers.BigNumber.from(1600);
  const l2GasLimit = ethers.BigNumber.from(23100);
  
  // Calculate gas costs
  const l1GasCost = l1GasLimit.mul(l1GasPrice);
  const l2GasCost = l2GasLimit.mul(l2GasPrice);
  
  // Add a buffer for price fluctuations (10%)
  const buffer = l1GasCost.add(l2GasCost).mul(10).div(100);
  
  // Total gas cost includes L1 fee, L2 gas, and buffer
  const totalCost = l1GasCost.add(l2GasCost).add(buffer);
  
  return {
    totalCost,
    gasLimit: l2GasLimit,
    gasPrice: l2GasPrice
  };
}
```

packages/dashboard/app/withdraw/utils/handlers.ts
awisniew207: Changed to `gte`

packages/dashboard/app/withdraw/utils/handlers.ts
awisniew207: Ideally the developer dashboard wouldn't use this at all. Since this is the only place that uses `LitNodeClient` right now, I think we can keep it until this page is moved to a user dashboard, where we can then set it up in the parent. Let me know if that works for you. I don't think it makes sense to move this to the parent if all of this code will be moved soon anyways

packages/dashboard/app/withdraw/utils/alchemyUtils.tsx
awisniew207: This has been done :+1:

packages/dashboard/app/withdraw/utils/handlers.ts
awisniew207: I've moved the LitNodeClient to it should only initialize once

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,277 @@
...
+      };
+    }
+
+    if (amount.lt(gasInfo.totalCost)) {
+      return {
+        success: false,
+        hash: '',
+        error: `Insufficient balance for gas fees.`
+      };
+    }
```
FedericoAmura, line 161:
```suggestion
    if (tokenBalance.lt(amount.add(gasInfo.totalCost))) {
      return {
        success: false,
        hash: '',
        error: `Insufficient balance to withdraw specified amount and pay for current gas fees.`
      };
    }
```

The second check is avoiding user to withdraw less than the gas they have to pay (might sound stupid but if they have that `tokenBalance` available after gas they should be able to do so)

Nonetheless, what you want to check is that with your balance you can pay for gas and retire the amount you want. Both things at the same time, together.
Having the checks separated might create edge cases where both conditions are true but the combination is not (ex. user wants to withdraw the 100% of their balance which is also more than the gas they will have to pay)

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,277 @@
...
+      gasPrice: ethers.utils.formatUnits(gasInfo.gasPrice, 'gwei') + ' gwei',
+      totalGasCost: ethers.utils.formatEther(gasInfo.totalCost) + ' ETH'
+    });
+
+    return await sendTransactionWithRetry({
+      pkpWallet,
+      txOptions,
+      provider,
+      initialGasPrice: gasInfo.gasPrice
+    });
```
FedericoAmura, line 184: Following on the comment above, this function might double the gas cost which makes the previous check not valid anymore unless it considered that price increase

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,277 @@
...
+      const estimatedGas = await tokenWithSigner.estimateGas.transfer(
+        recipientAddress,
+        amount
+      );
+
+      // Add 10% buffer to estimated gas for token transfers
+      gasLimit = estimatedGas.mul(110).div(100);
+      console.log('Estimated gas for token transfer:', gasLimit.toString());
+    } catch (err) {
+      console.warn('Failed to estimate gas for token transfer, using default:', err);
```
FedericoAmura, line 230: There is no default gas...

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,245 @@
...
+async function initLitNodeClient() {
+  if (!isLitNodeClientInitialized) {
+    litNodeClient = new LitNodeClient({
+      litNetwork: SELECTED_LIT_NETWORK
+    });
+    await litNodeClient.connect();
+    isLitNodeClientInitialized = true;
+  }
+  return litNodeClient;
+}
```
FedericoAmura, line 21: Note: On the migration to vite let's move these things to providers, contexts and hooks. Check the [doc here](https://react.dev/learn/passing-data-deeply-with-context)

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,245 @@
...
+  if (!ethAddress) return;
+
+  try {
+    setLoading(true);
+    const ethBalance = await fetchEthBalance(ethAddress);
+    const erc20TokenBalances = await fetchERC20TokenBalances(ethAddress);
+
+    const balances = [ethBalance.balances[0],...erc20TokenBalances.balances];
+
+    if (erc20TokenBalances.success && ethBalance.success) {
```
FedericoAmura, line 115:
```suggestion
    if (erc20TokenBalances.success && ethBalance.success) {
        const balances = [...ethBalance.balances, ...erc20TokenBalances.balances];
```
`balances` is only used inside that block. And the existence of `[0]` assumes both `ethBalance.success: true` and that the `ethBalance.balances` array is not empty (just destructuring is looser so it won't throw when the array is empty for example or would allow multiple native balances for when we do this multichain)

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  if (selectedToken) {
+    if (selectedToken.symbol === 'ETH') {
+      const parsedAmount = ethers.utils.parseUnits(withdrawAmount, selectedToken.decimals);
+      if (parsedAmount.gte(selectedToken.rawBalance)) {
+        showStatus(`You have entered an amount greater than or equal to your balance. Please enter a smaller amount.`, 'warning');
+        return;
+      }
+    }
+  }
```
FedericoAmura, line 174: This should also check for the required gas. The check should be the same as in other parts of code

nit: `selectedToken` is guaranteed to have a value at this point. First `if` is pointless

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,245 @@
...
+      rpc: BASE_MAINNET_RPC
+    })
+
+    await pkpWallet.init();
+
+    const amount = ethers.utils.parseUnits(withdrawAmount, selectedToken.decimals);
+
+    let transactionResult;
+
+    if (selectedToken.symbol === 'ETH') {
```
FedericoAmura, line 194: This might be a big refactor at this point and possibly better a follow up. But treating the native coin as an token pollutes the code with so much of these ifs that it is just not worth it

The benefit of merging them seems reasonable when thinking their display is similar (which really is not because we always display the native token) but they have so many logical differences (ones pays gas and is a native transfer the other can be maxed and triggers a contract) that make the tradeoff not worth it

Just wanted to write that. We don't need to fix it now as it is expensive, it works, we need this shipped and we are going to refactor a bunch of stuff already

packages/dashboard/app/withdraw/WithdrawForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+      hideToken, 
+      unhideToken, 
+      setSelectedToken, 
+      setWithdrawAmount
+    );
+  };
+
+  // Wrapper for submit
+  const onSubmit = (e: React.FormEvent) => {
+    handleSubmit(
```
FedericoAmura, line 132: Both these functions are passing the event to a function that is not in "component land" and should not have to care about handling events.

The call to `e.preventDefault()` or `e.stopPropagating()` should be here and not have the event passed to the functions

packages/dashboard/components/withdraw/WalletInfo.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const WalletInfo: React.FC<WalletInfoProps> = ({ ethAddress }) => {
+  return (
+    <div className="wallet-info">
+      <h3>Wallet Information</h3>
+      <div className="wallet-address">
+        <strong>EVM Address:</strong> {ethAddress || 'Not available'}
+      </div>
+      <div className="wallet-network">
+        <strong>Network:</strong> Base Mainnet
```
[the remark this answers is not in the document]
awisniew207: Right, we're just doing Base Mainnet right now because we need to get this to users

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,277 @@
...
+      console.log('Retrying with increased gas price:', ethers.utils.formatUnits(txOptions.gasPrice, 'gwei') + ' gwei');
+
+      const retryTx = await pkpWallet.sendTransaction(txOptions);
+      const retryReceipt = await provider.waitForTransaction(retryTx.hash, 1);
+
+      return {
+        success: retryReceipt.status === 1,
+        hash: retryTx.hash,
+        receipt: retryReceipt
+      };
```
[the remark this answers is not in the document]
awisniew207: I definitely agree, but knowingly did this as a temporary fix for Base Mainnet in case a user encounters this issue. If they get the replacement fee too low error and can't set their gas price, they have no other way of submitting a transaction right now. This will be updated with multichain support

packages/dashboard/app/withdraw/utils/transactionService.ts
awisniew207: Sorry, accidentally left that in there. There used to be

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,188 @@
...
+      success: receipt.status === 1,
+      hash: tx.hash,
+      receipt
+    };
+  } catch (error: any) {
+    if (error.message && error.message.includes('insufficient funds for intrinsic transaction cost')) {
+      return {
+        success: false,
+        hash: '',
+        error: 'Insufficient funds for transaction. Please add more ETH to your wallet.'
```
FedericoAmura, line 60:
```suggestion
        error: 'Insufficient funds to pay for transaction. Please add more funds to your wallet.'
```

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,188 @@
...
+      pkpWallet,
+      txOptions,
+      provider,
+    });
+  } catch (error: any) {
+    console.error('Error sending ETH transaction:', error);
+    return {
+      success: false,
+      hash: '',
+      error: error.message || 'Failed to send ETH'
```
FedericoAmura, line 106:
```suggestion
      error: error.message || 'Failed to send native asset'
```

Because we are multichain now we cannot talk about Eth anymore. This should at least be changed in the user facing strings

packages/dashboard/app/withdraw/utils/transactionService.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    const ethBalance = await pkpWallet.getBalance();
+    const gasCost = gasLimit.mul(gasPrice);
+
+    if (ethBalance.lt(gasCost)) {
+      return {
+        success: false,
+        hash: '',
+        error: `Insufficient ETH for gas fees. Need ${ethers.utils.formatEther(gasCost)} ETH for gas.`
```
FedericoAmura, line 171:
```suggestion
        error: `Insufficient native balance for gas fees. Need ${ethers.utils.formatEther(gasCost)} native coin for gas.`
```

packages/dashboard/app/withdraw/utils/handlers.ts
```diff
@@ -0,0 +1,147 @@
...
+        transactionResult = await sendTokenTransaction({
+          pkpWallet,
+          tokenDetails: token,
+          amount,
+          recipientAddress: withdrawAddress,
+          provider
+        });
+        showStatus(`Detected token: ${symbol}`, 'info');
+      } catch (error) {
+        showStatus('Could not fetch token details.', 'error');
```
FedericoAmura, line 113:
```suggestion
        showStatus('Could not fetch token details. Check token address and try again.', 'error');
```
Good thing is that we get this error when trying to use a non ERC20 address

FedericoAmura (approved):
Let's ship this!

PR #80, opened 2025-04-22
Dashboard migration to vite

FedericoAmura:
# Description

This PR adds a new `app-dashboard` which is a migration of the current `dashboard` but based in Vite runtime instead of Next.js

# Motivation

We were not using any of the features of Next.js and we were getting random 500 errors (server) when the full app was client side. In other words, we were not getting any benefit from next but getting issues from it

# Followup

**AFTER** this PR has been merged, #83 MUST be merged too to remove the dashboard this PR migrates. The migration is done this way so both dashboards can be deployed and coexist for some time while reviewing, migrating deployment, domain and all the necessary config

# Changes
- There is a new [vite based dashboard](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard) nearly identical to the [previous one](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/dashboard) (remember followup above). No more SSR > No more 500 (server) errors neither dirty hacks to avoid hydration mismatch or client side only rendering
- Routing is still directory based [here](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard/src/pages) but without the layout files composition. Check how the page just defines itself and then exports it with the surrounding stuff it needs using [components utility functions](https://github.com/LIT-Protocol/Vincent/blob/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard/src/utils/components.tsx)
- There are no more components in the `pages` directory (one could query a component by its url). They have been moved mostly [here](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard/src/components)
- Layout has been separated into app [dashboard] and user [dashboard] ones [here](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard/src/components/layout). Nearly identical for now; app devs use a Header but users didn't need it and could get confused with the extra EOA login
- Providers have been separated too for app devs and users [here](https://github.com/LIT-Protocol/Vincent/tree/feature/drel-588-nextjs-500-error-vite-migration/packages/app-dashboard/src/providers). No more stytch provider for app devs and no more rainbow or wagmi providers for users. Less moving parts that can fail
[...]

PR #83, opened 2025-04-23
feat: remove old dashboard

FedericoAmura:
# Description

This PR removes the old dashboard after it has been migrated to vite in #80 
It also reacomodates the dependencies that were all originally in the root package.json to clarify which packages should include each dep and which ones can/should be shared


awisniew207 (approved):
lgtm!

PR #89, opened 2025-04-28
feat: `vincent-tool-sdk` author functionality for tools and policies

MaximusHaximus:
Initial cut of `vincent-tool-sdk`
- Renamed package
- Implemented `createVincentPolicy`, `createVincentTool`, and `createVincentToolPolicy` methods that provide strong type inference for the lifecycle methods you provide on the object that is passed into the functions, based on the schemas you provide.
- Implemented a context object for policies that provide helper methods for `allow` and `deny` actions
- Implemented a context object for tools that provide helper methods for `succeed` and `fail` actions
- The context object for the Tool contains the `policiesContext` object, which will contain both the evaluation results and the `commit()` handler for any policies that were evaluated.
- The policiesContext object during `precheck` has no `commit()` handlers, just results.
- policiesContext objects are keyed on the _package name of the policy_ rather than the `ipfsCid` (yay!)
- Added extensive typescript type verification files to the package so we can be sure we don't break inference in the future.


spacesailor24 (approved):

spacesailor24 (changes requested):
As per our call, we need Policy `userParamsSchema` to be optional, and we were also considering making `userPkpTokenId` a required base param for both Tool and Policy toolParams

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
```diff
@@ -0,0 +1,245 @@
...
+  supportedPolicies: Policies;
+  precheck: (
+    params: z.infer<ToolParamsSchema>,
+    policyEvaluationResults: VincentPolicyEvaluationResults<Policies>,
+  ) => Promise<unknown>;
+  execute: (
+    params: z.infer<ToolParamsSchema>,
+    policyEvaluationResults: VincentPolicyEvaluationResults<Policies>,
+  ) => Promise<unknown>;
+}
```
spacesailor24, line 245: Don't we want the return types to have some structure? e.g.

```tsc
{ success: true, result: unknown }

{ success: false, reason: string }
```

or something similar?

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
```diff
@@ -0,0 +1,245 @@
...
+  policyDef: PolicyDefType;
+  toolParameterMappings: Partial<{
+    [K in keyof z.infer<ToolParamsSchema>]: keyof z.infer<
+      PolicyDefType['toolParamsSchema']
+    >;
+  }>;
+};
+
+// Policy evaluation results with proper typing for commit
+export type VincentPolicyEvaluationResults<
```
spacesailor24, line 200: I think this is also missing the Policy commit execution results

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
MaximusHaximus: :+1: Pushed up proper support for schema-driven return values for precheck and execute on the VincentTool -- sorry for the oversight. I also added the same context pattern to these methods and fleshed out the playground into a set of files that will actually fail to compile if we break any of the inference rules.

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
MaximusHaximus: We need to add policy commit execution results into the _tool_ response, since they aren't actually part of the evaluation lifecycle. 🤔  I'll look into that tomorrow

packages/vincent-tools/vincent-tool-sdk/src/lib/playground.ts
```diff
@@ -0,0 +1,342 @@
...
+  actionType: z.string(),
+});
+const userParams1Schema = z.object({ allowedTargets: z.array(z.string()) });
+const commitParams1Schema = z.object({ confirmation: z.boolean() });
+
+// Define the result schemas for policy1
+const policy1EvalAllowResult = z.string(); // Result when allowed
+const policy1EvalDenyResult = z.object({ reason: z.string() }); // Result when denied
+const policy1CommitAllowResult = z.object({ success: z.boolean() });
+const policy1CommitDenyResult = z.object({ errorCode: z.number() });
```
FedericoAmura: Would it make sense to add this playground to the readme or docs? Seems like something app devs will want to do

Also, can we have more descriptive names? What does `policy1` do and what it has different from `policy2` and `policy3`?

packages/vincent-tools/vincent-tool-sdk/src/lib/playground.ts
```diff
@@ -0,0 +1,342 @@
+import z from 'zod';
```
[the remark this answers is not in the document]
MaximusHaximus: True, you ain't lyin -- fixed!

packages/vincent-tools/vincent-tool-sdk/src/lib/vincent-policy/index.ts
```diff
@@ -0,0 +1,98 @@
+import z from 'zod';
```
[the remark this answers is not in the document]
MaximusHaximus: <3

PR #92, opened 2025-04-30
Init Vincent Tool and Policy Handlers

spacesailor24:


packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+        const allOnChainPolicyParams = await _getAllOnChainPolicyParams(args);
+
+        // We exit early here because !allOnChainPolicyParams.isPermitted means appDelegateeAddress
+        // is not permitted to execute toolIpfsCid for the Vincent App on behalf of the agentWalletPkpTokenId
+        // and no further processing is needed
+        if (!allOnChainPolicyParams.isPermitted) {
+            throw new Error(`App Delegatee: ${args.appDelegateeAddress} is not permitted to execute Vincent Tool: ${args.toolIpfsCid} for App ID: ${allOnChainPolicyParams.appId.toString()} App Version: ${allOnChainPolicyParams.appVersion.toString()} using Agent Wallet PKP Token ID: ${args.agentWalletPkpTokenId}`);
+        }
+
+        const onChainPolicyParams = allOnChainPolicyParams.policies.find(
```
MaximusHaximus, line 34: Now that we're running this per-policy, can we use the contract function you mentioned that just returns params for this one policy, rather than relying on a .find() to select it? We're getting a lot more data here than we need.

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+            (policy) => policy.policyIpfsCid === args.policyIpfsCid
+        );
+
+        if (args.policyUserParamsSchema) {
+            if (!onChainPolicyParams) {
+                // All user Policy params can be optional, so if no parameters were set on-chain,
```
MaximusHaximus, line 40: 💖 If this happens, we can assume that the Zod schema either has no required properties, or it has some required properties but the policy developer failed to mark them as required in the JsonSchema / uiSchema

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+}): Promise<AllOnChainPolicyParams> => {
+    try {
+        const VINCENT_CONTRACT_ABI = [
+            `function validateToolExecutionAndGetPolicies(address delegatee, uint256 pkpTokenId, string calldata toolIpfsCid) external view returns (tuple(bool isPermitted, uint256 appId, uint256 appVersion, tuple(string policyIpfsCid, tuple(string name, uint8 paramType, bytes value)[] parameters)[] policies) validation)`,
+        ];
+
+        const vincentContract = new ethers.Contract(
+            args.vincentContractAddress,
+            VINCENT_CONTRACT_ABI,
+            new ethers.providers.JsonRpcProvider(
```
MaximusHaximus, line 75: `StaticJsonRpcProvider` instead

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+    yellowstoneRpcUrl: string,
+    vincentContractAddress: string,
+    appDelegateeAddress: string,
+    agentWalletPkpTokenId: string,
+    toolIpfsCid: string,
+    policyIpfsCid: string,
+    policyUserParamsSchema?: z.infer<VincentPolicyDef['userParamsSchema']>
+}): Promise<z.infer<VincentPolicyDef['userParamsSchema'] | undefined>> => {
+    try {
+        const allOnChainPolicyParams = await _getAllOnChainPolicyParams(args);
```
MaximusHaximus, line 25: This method throws its own very descriptive error and we don't need to wrap it -- move outside of `try{}` IMO

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        // We exit early here because !allOnChainPolicyParams.isPermitted means appDelegateeAddress
+        // is not permitted to execute toolIpfsCid for the Vincent App on behalf of the agentWalletPkpTokenId
+        // and no further processing is needed
+        if (!allOnChainPolicyParams.isPermitted) {
```
MaximusHaximus, line 30: Same as above, move outside of `try{}` IMO, we don't want this error wrapped

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+                const decodedPolicyParams = abiDecodePolicyParameters({ params: onChainPolicyParams.parameters });
+                return parseOnChainPolicyParams(args.policyUserParamsSchema.parse(decodedPolicyParams));
+            }
+        } else {
+            if (!onChainPolicyParams) {
+                return;
+            } else {
+                throw new Error(`Agent Wallet PKP Token ID: ${args.agentWalletPkpTokenId} has registered on-chain Policy parameters for Vincent App: ${allOnChainPolicyParams.appId.toString()} App Version: ${allOnChainPolicyParams.appVersion.toString()} Vincent Policy: ${args.policyIpfsCid} Vincent Tool: ${args.toolIpfsCid} but no userParamsSchema was defined by the Policy`);
+            }
+        }
```
MaximusHaximus, line 54: This type of double nesting can be really tricky to grok, and having error cases at the very end adds to cognitive overhead

Let's flatten out this nesting and throw _early_ for the error case. By running a few conditional checks in a row, we make it a lot easier to reason through the logic... something like this (warning: I didn't run this or write it in an IDE)
```suggestion
        if(onChainPolicyParameters && !args.policyUserParamsSchema) {
          throw new Error(`Agent Wallet PKP Token ID: ${args.agentWalletPkpTokenId} has registered on-chain Policy parameters for Vincent App: ${allOnChainPolicyParams.appId.toString()} App Version: ${allOnChainPolicyParams.appVersion.toString()} Vincent Policy: ${args.policyIpfsCid} Vincent Tool: ${args.toolIpfsCid} but no userParamsSchema was defined by the Policy`);
        }
        
        if (!onChainPolicyParams && args.policyUserParamsSchema) {
          // All user Policy params can be optional, so if no parameters were set on-chain,
          // we want to validate an empty object (i.e. no parameters) against the userParamsSchema
          // to validate that there are no required Policy params
          return parseOnChainPolicyParams(args.policyUserParamsSchema.parse({}));
        }
        
        // Standard case; we have on chain params and we have a schema to verify!
        if(args.policyUserParamsSchema && onChainPolicyParams) {
          const decodedPolicyParams = abiDecodePolicyParameters({ params: onChainPolicyParams.parameters });
          return parseOnChainPolicyParams(args.policyUserParamsSchema.parse(decodedPolicyParams));
        }
        
        // Standard case; we have no on chain parameters and no schema!
        return;
```

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+    } catch (error) {
+        throw new Error(`Error getting on-chain policy parameters (getOnChainPolicyParams): ${error instanceof Error ? error.message : String(error)}`);
+    }
```
MaximusHaximus, line 57: Given how explicit you made your errors for all of the child functions here, and with the changes i mentioned above, I think we can :hocho: this function-wide try/catch; it actually only serves to obfuscate the real error now that all of your child functions throw very explicit errors. :+1: Nice work

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
```diff
@@ -0,0 +1,97 @@
...
+    } catch (error) {
+        throw new Error(`Error getting on-chain policy parameters from Vincent contract: ${args.vincentContractAddress} using App Delegatee: ${args.appDelegateeAddress} and Agent Wallet PKP Token ID: ${args.agentWalletPkpTokenId} and Vincent Tool: ${args.toolIpfsCid}: ${error instanceof Error ? error.message : String(error)}`);
+    }
+}
+
+const parseOnChainPolicyParams = (onChainPolicyParams: Record<string, EthersAbiDecodedValue>, policyUserParamsSchema?: z.infer<VincentPolicyDef['userParamsSchema']>) => {
+    try {
+        return policyUserParamsSchema.parse(onChainPolicyParams);
+    } catch (error) {
+        const errorMessage = error instanceof z.ZodError ? formatZodError(error) : error instanceof Error ? error.message : String(error);
```
MaximusHaximus, line 94: formatZodError returns an object w/ message and other data -- we want that object to be converted to an error `message` string for use here - don't just rely on the object being stringified.

packages/vincent-tools/vincent-tool-sdk/src/lib/lit-actions/vincent-policy-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+            });
+        }
+    }
+}
+
+const parseToolParams = (toolParams: z.infer<VincentPolicyDef['toolParamsSchema']>, toolParamsSchema: z.ZodType<any, any, any>) => {
+    try {
+        return toolParamsSchema.parse(toolParams);
+    } catch (error) {
+        const errorMessage = error instanceof z.ZodError ? formatZodError(error) : error instanceof Error ? error.message : String(error);
```
MaximusHaximus, line 62: Same as above re: we want `formatZodError` to give us a message _string_

packages/vincent-tools/vincent-tool-sdk/src/lib/lit-actions/vincent-policy-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+                policyIpfsCid: policyDef.ipfsCid,
+                policyUserParamsSchema: policyDef.userParamsSchema,
+            });
+
+            const evaluateResult = await policyDef.evaluate({ toolParams: parsedToolParams, userParams: onChainPolicyParams });
+
+            Lit.Actions.setResponse({
+                response: JSON.stringify({
+                    ...evaluateResult,
+                    ipfsCid: policyDef.ipfsCid,
```
MaximusHaximus, line 43: Can we verify that the current LIT Action ipfsCid matches the policyDef ipfsCid?

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
```diff
@@ -243,3 +244,31 @@ export interface VincentToolDef<
     policyEvaluationResults: VincentPolicyEvaluationResults<Policies>,
   ) => Promise<unknown>;
 }
+
+export interface PolicyParameter {
```
MaximusHaximus, line 248: Since package consumers should only get the post-decoding and parsing values, we need to make sure that these aren't exported from the package. 🤔 Maybe we can move them into a `types.ts` file inside of the `Vincent-policy`  instead of it being in our root types.ts

MaximusHaximus (changes requested):
Added a review re: the policy stuff -- I know the Tool Handler is still in progress so I didn't review that.

packages/vincent-tools/vincent-tool-sdk/src/lib/get-onchain-policy-params.ts
spacesailor24: There isn't a function to return a single policies' parameter, we only have the single read function that returns all the  of the registered policy params for all the tools for an App for a specific Agent Wallet

packages/vincent-tools/vincent-tool-sdk/src/lib/lit-actions/vincent-policy-handler.ts
spacesailor24: It would be best to just use `LitAuth.actionIpfsIds[0]` right? Use the actual IPFS CID that's being executed as reported by the Lit node

packages/vincent-tools/vincent-tool-sdk/src/lib/types.ts
spacesailor24: Moved them to `lit-actions/types.ts`

spacesailor24:
Superseded by #97

PR #95, opened 2025-05-01
User Dashboard

awisniew207:
New PR for the User Dashboard. 

Duplicated files with slight modifications. These will be set to replace the existing versions upon removal of this functionality from the Consent Page. These shouldn't require much review, since their changes are truly minimal:
- `AuthenticatedConsentForm` -> `UserAuthenticatedConsentPage`
- `VersionUpgradePrompt` -> `UserVersionUpgradePrompt`
- `useAppPermissionCheck` -> `useUserAppPermissionCheck`
- `index` -> `consent`
  - Unified the Consent flows based on their location, can be changed with a flag
  - One signs you in automatically, the other prompts first
    
These files account for 2/3 of the PR length.

PR #96, opened 2025-05-01
Better Route Organization, Withdraw Page Simplification

awisniew207:
- Better organization of routes, removal of the `wrap function`
- Simplified the Withdraw Page to consume existing authentication, not create a new `ConsentView`
- Better auth handling for User Dashboard Routes. Any unauthenticated users will be redirected to `/user` to authenticate again


packages/app-dashboard/src/components/consent/pages/consent.tsx
```diff
@@ -302,7 +302,6 @@ export default function ConsentView({ isUserDashboardFlow = false }: ConsentView
     if (isUserDashboardFlow && validatedSessionSigs && authInfo?.userPKP) {
       // Redirect to /apps page
       navigate('/user/apps');
-      return null;
```
FedericoAmura, line 305: This function is expected to return page content. On this case (which should be really a hook, let's fix that on the parent PR. I'll add the comment there for this) we should still return something. For example `<></>`

The concept is that this function takes some values and returns the needed view of it, even if it is an "empty" view

packages/app-dashboard/src/components/layout/UserHeader.tsx
```diff
@@ -49,10 +49,14 @@ export default function UserHeader({ backButton, title, onSignOut }: UserHeaderP
           )}
 
           {!isRootPath && (
-            <Button onClick={handleSignOut} variant="outline" size="sm">
-              <LogOut className="mr-2 h-4 w-4" /> Sign Out
+            <Button onClick={() => navigate('/user/withdraw')} variant="outline" size="sm">
+              <CreditCard className="mr-2 h-4 w-4" /> Withdraw
```
FedericoAmura, line 53: Let's use https://lucide.dev/icons/banknote-arrow-down

Credit card could be interpreted as deposit, buy, sell, anything

packages/app-dashboard/src/components/user/AuthGuard.tsx
```diff
@@ -0,0 +1,25 @@
...
+    );
+  }
+
+  if (!authInfo?.userPKP || !authInfo?.agentPKP || !sessionSigs) {
+    return <Navigate to="/user" replace />;
+  }
+
+  // Return null if authenticated, allowing the protected content to render
+  return null;
+}
```
FedericoAmura, line 25: Some things with this hook

- The response should be as consistent as possible. In functions, not only in hooks really. In this case we are returning 3 different things: a loading component, a navigation component that will change app state or a `null`
- Caller will have a very hard time trying to understand if the returned component is a loading (which they might want to overload) or a navigate (which they cannot overload without breaking the goal of calling this)
- Caller has to be aware that what it receives might be, or not, something it CAN+MUST render (when you want to return no component return `<></>` (it is kind of the component equivalent of `null`))
- React router discourages `Navigate` over `useNavigate`. Navigate is a workaround where you can't use the hook

A much better implementation
```suggestion
export function useAuthGuard() {
  const { authInfo, sessionSigs, isProcessing } = useReadAuthInfo();
  const navigate = useNavigate();

  const userIsAuthed = authInfo?.userPKP && authInfo?.agentPKP && sessionSigs;

  if (!isProcessing && !userIsAuthed) {
    navigate(`/user`, { replace: true });
  }
  
  return isProcessing;
}
```

This implementation does just 1 thing: redirect if needed. And lets caller know it is in process of doing that. It doesn't force the caller to use a loading component nor returns different type of things on the same variable

And then in components that need this just add a `useAuthGuard()` at the top, If components wants to show a loading, skeleton or whatever, then they can decide it and customize it any way they want (there is a loading component already btw, let's stop copying this code)

packages/app-dashboard/src/pages/user-dashboard/app-details.tsx
```diff
@@ -2,9 +2,15 @@ import { Helmet } from 'react-helmet';
...
 import UserAuthenticatedConsentForm from '@/components/consent/components/UserAuthenticatedConsentForm';
 import UserHeader from '@/components/layout/UserHeader';
+import { useAuthGuard } from '@/components/user/AuthGuard';
 
 export default function AppDetailsPage() {
   const { authInfo, sessionSigs } = useReadAuthInfo();
+  const authGuardElement = useAuthGuard();
+
+  if (authGuardElement) {
+    return authGuardElement;
```
FedericoAmura, line 12: Following the comment above. If the guard is pending, then this page should decide what to do. Almost always showing a `<Loading />` until the guard resolves and redirects or says it finished without issue

This way the guard has also the responsibility of telling the caller what to BLINDLY render.

FedericoAmura (commented):
I would prefer we merge this and check everything together in #95 . Comments still apply but checking things there that might be changed here makes the review more complicated

packages/app-dashboard/src/components/consent/pages/consent.tsx
awisniew207: Temporarily changed to:
```
return <></>;
```

packages/app-dashboard/src/components/layout/UserHeader.tsx
awisniew207: Changed to:
```
<Button onClick={() => navigate('/user/withdraw')} variant="outline" size="sm">
  <BanknoteArrowDown className="mr-2 h-4 w-4" /> Withdraw
</Button>
```

PR #102, opened 2025-05-12
Vincent APIs

awisniew207:
App APIs - Posted here
Tool APIs - Posted here
Policy APIs - Posted here

All APIs match the off-chain spec


packages/app-dashboard/temp-apis/app.yaml
```diff
@@ -0,0 +1,531 @@
...
+          example: 2
+        toolPackageName:
+          type: string
+          example: "@vincent/foo-bar"
+        toolVersion:
+          type: string
+          example: "1.0.0"
+        #appVersionIdentity:
+        #  type: string
+        #  example: "5@2" <- Why not include version number in the overall identity?
```
awisniew207, line 501: This was in the spec, but I've commented it since I'm not completely sure we need it

packages/app-dashboard/temp-apis/app.yaml
```diff
@@ -0,0 +1,999 @@
...
+          example: "QmdoY1VUxVvxShBQK5B6PP2jZFVw7PMTJ3qy2aiCARjMqo"
+          
+    ICreateToolVersionDef:
+      type: object
+      properties:
+        changes:
+          type: string
+          example: "Extra foo on the bar!" # Is changes the only property we need?
+                                           # I'm not sure how much is provided by our BE
+
```
awisniew207, line 989: Other point I wasn't too sure about

packages/vincent-rest-api/src/utils/generateOpenAPIClient.ts
```diff
@@ -0,0 +1,22 @@
+import { exec } from 'child_process';
+import * as path from 'path';
+
+const yamlPath = path.join(__dirname, '../api/generated-openapi.yaml');
+const outputDir = path.join(__dirname, '../common/api');
+
+const command = `npx openapi --input ${yamlPath} --name Api --postfixServices Api --indent 2 --output ${outputDir}`;
```
MaximusHaximus, line 7: If we do need to run openapi from the command line, let's just run the `pnpx` command directly from our `package.json` without this wrapper file, using relative paths from the package.json for the locations of the files. I don't think there's anything in this file that justifies a TS wrapper

packages/vincent-rest-api/project.json
```diff
@@ -0,0 +1,30 @@
...
+  "name": "@lit-protocol/vincent-rest-api",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "packages/vincent-rest-api",
+  "projectType": "application",
+  "targets": {
+    "generate-api": {
+      "executor": "nx:run-commands",
+      "options": {
+        "cwd": "packages/vincent-rest-api",
+        "command": "ts-node src/utils/generate-openapi.ts && ts-node src/utils/generateOpenAPIClient.ts"
```
MaximusHaximus, line 11: 1. If this command is defined in the package.json, then I think that this entry doesn't actually need to exist in the `project.json` because commands in `package.json` _are_ nx targets (implicitly)
2. If we do need this here, use `nx:run-script` instead, where the target is the script name in the package.json to avoid duplicating the task definition. The cwd should be automatically inferred to be the location of the package.json so don't define it here

packages/vincent-rest-api/src/tests/integration.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      if ('appId' in response) {
+        expect(response).toHaveProperty('identity');
+        expect(response).toHaveProperty('appId');
+        expect(response.name).toBe(app.name);
+        expect(response.description).toBe(app.description);
+
+        appIdentity = response.identity;
+        console.log('Created app with identity:', appIdentity);
+      } else {
+        fail('Expected AppDef response but got Error');
```
MaximusHaximus, line 50: :hocho: this `fail()` thing from all of the tests -- it's a kludge.  The output for failures like this will be very poor. Also, this  check is just extra code -- it is wrapping an expect() right below it that actually is doing things correctly (using .toHaveProperty()). Assertions should be explicit and use `expect()` directly on the thing that you're asserting against.

packages/vincent-rest-api/src/tests/integration.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        description: 'An app for integration testing',
+        contactEmail: 'test@example.com',
+        appUserUrl: 'https://example.com',
+        logo: 'base64encodedlogo',
+        redirectUris: ['https://example.com/callback'],
+        deploymentStatus: ICreateAppDef.deploymentStatus.DEV,
+        managerAddress: '0xa723407AdB396a55aCd843D276daEa0d787F8db5',
+      };
+
+      const response = await apiClient.app.createApp(app);
```
MaximusHaximus, line 39: ```expect(apiClient.app......).resolves.toMatchObject({ name: app.name, description: app.description, appId: expect.any(Number) })```

packages/vincent-rest-api/src/tests/integration.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    test('create app - invalid schema', async () => {
+      const invalidApp = {
+        name: 'Invalid Test App',
+      };
+
+      try {
+        await apiClient.app.createApp(invalidApp as any);
+        fail('Expected API call to fail but it succeeded');
+      } catch (e: any) {
+        if (e instanceof ApiError) {
```
MaximusHaximus, line 65: This is a lot of boilerplate -- don't test for errors like this, it's very brittle, hard to reason about, and verbose. What you want is something like

```
const result = await expect(apiClient.app....).rejects.toThrow();
```

MaximusHaximus (changes requested):
Initial comments / changes added in-line.

Also, let's get all of the generated code to live in `/src/generated/**/*`.  Then add that path to the .gitattributes so that Github knows not to show the file contents of generated side-effects in PRs:

https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

packages/app-dashboard/temp-apis/app.yaml
MaximusHaximus: Yes, I think right now we only allow `changes`

packages/app-dashboard/temp-apis/app.yaml
MaximusHaximus: Entity identities should  only exist in the DB backend 🔪 The version and package name should be enough for the client side I think

packages/app-dashboard/package.json
```diff
@@ -40,6 +41,10 @@
     "eslint-plugin-react-hooks": "^5.2.0",
     "eslint-plugin-react-refresh": "^0.4.19",
     "globals": "^16.0.0",
+    "http-server": "^14.1.1",
```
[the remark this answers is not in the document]
awisniew207: Removed! These were there since I initially started this work in that package, before migrating it to the current

packages/vincent-rest-api/docs/api.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <script type="module" src="https://unpkg.com/rapidoc/dist/rapidoc-min.js"></script>
+</head>
+
+<body>
+  <rapi-doc spec-url="/packages/temp-apis/src/api/generated-openapi.yaml" theme="light"> </rapi-doc>
```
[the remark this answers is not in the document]
awisniew207: Not generated, but not update from when I moved some files around for better organization. Pushed a change so that running `http-server` from the root will allow you to see this for better UI 
image

packages/vincent-rest-api/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@lit-protocol/vincent-rest-api",
+  "version": "1.0.0",
+  "description": "Integration tests for Vincent Registry API",
+  "scripts": {
+    "generate-openapi": "npx ts-node src/utils/generate-openapi.ts",
+    "generate-client": "npx ts-node src/utils/generateOpenAPIClient.ts",
+    "generate-api": "npx ts-node src/utils/generate-openapi.ts && npx ts-node src/utils/generateOpenAPIClient.ts",
```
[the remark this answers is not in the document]
awisniew207: Changed to:
```
    "generate-openapi": "pnpx ts-node src/utils/generate-openapi.ts",
    "generate-client": "pnpx ts-node src/utils/generateOpenAPIClient.ts",
    "generate-api": "run-s generate-openapi generate-client",
```

packages/vincent-rest-api/src/utils/generateOpenAPIClient.ts
awisniew207: Changing to:

```
"generate-client": "pnpx openapi-typescript-codegen --input src/api/generated-openapi.yaml --name Api --postfixServices Api --indent 2 --output ../common/api",
```

packages/vincent-rest-api/src/tests/server.ts
```diff
@@ -0,0 +1,325 @@
+import express, { Request, Response } from 'express';
```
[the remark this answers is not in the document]
awisniew207: Deleting this file

packages/vincent-rest-api/project.json
awisniew207: I don't see why it would be needed at the moment. Do we actually need the `project.json` here at all?

awisniew207:
Adding:
```
packages/vincent-rest-api/src/generated/**/* linguist-generated=true
```

To the `.gitattributes`

packages/vincent-rest-api/src/tests/integration.test.ts
awisniew207: Yeah the integration tests are a heavy WIP right now.... I'd recommend not looking at them until the end :cold_sweat:

awisniew207:
Currently missing one route, which is `GET /apps`:

Lists apps in an arbitrary order (newest first?) and paginated

I'm not entirely sure what we'd like the structure of the returned paginated object to be

PR #104, opened 2025-05-16
Vincent MCP Server

FedericoAmura:
# Description

This PR adds:
- a Vincent App -> MCP Server (without transport aka. interface) to the Vincent SDK. This transformer is used to convert any Vincent App into an MCP Server that can be extended with more tools or other resources and then exposed with the required interface that the LLM client needs
- a new package that is an actual MCP Server hosting solution with STDIO and HTTP transports (aka. interfaces) to use in local and remote mode by developers (not end users, that is a future version). This package also adds tools to query delegators, native balances, erc20 balances and erc20 allowances. This handles most of the required work that any Vincent App developer will have to do for setup only requiring it to provide his delegatee PK and a Vincent App definition file to define and expose the tools. A following step will be to add an `npx` command in this package so devs won't even need to get this source code

To use and test the MCP server with Claude Desktop (or any other compatible LLM client) the following config can be provided (with the adjusted config)
```
{
    "mcpServers": {
        "uniswapSwap": {
          "command": "npx",
          "args": [
            "-y",
            "tsx",
            "--env-file=/<ADJUST_DIRECTORY>/vincent-mcp/.env",
            "/<ADJUST_DIRECTORY>/vincent-mcp/src/stdio.ts"
          ]
        }
    }
}
```

Added Doc own section with references to OpenAI AgentKit and the full MCP Server using STDIO or HTTP transports
Screenshot from 2025-05-22 21-11-16
Screenshot from 2025-05-22 21-11-23
Screenshot from 2025-05-22 21-11-35

And API Doc
Screenshot from 2025-05-22 21-24-10


packages/mcp/.env.example
```diff
@@ -0,0 +1,7 @@
+HTTP_PORT=
+
+PUBKEY_ROUTER_DATIL_CONTRACT=0xF182d6bEf16Ba77e69372dD096D8B70Bc3d5B475
```
glitch003, line 3: This is fine for now, but just want to point out we should make this easier.  They should be able to point to a network like "naga-test" and we should be able to resolve all these items using the JS SDK, by reading from the ContractResolver.sol contract.  We should add the Vincent contract to it.  We don't technically need to modify the contract to do this - it supports resolving any contract name out of the box, but we do have to set up the deploy flow to include the vincent contract in the resolver automatically during contract deploy.

glitch003 (approved):
looks great, thanks!

packages/mcp/.env.example
DashKash54: Why do we even need this? Is something Vincent related missing from the Vincent contracts? We should use the Vincent Contracts SDK which @spacesailor24  is working on

packages/mcp/.env.example
```diff
@@ -0,0 +1,7 @@
...
+VINCENT_DATIL_CONTRACT=0x78Cd1d270Ff12BA55e98BDff1f3646426E25D932
```
DashKash54, line 4: Same for this ^

packages/mcp/README.md
```diff
@@ -0,0 +1,92 @@
+# Vincent MCP Server
+
+This package contains a Vincent App Server that can be used to serve Vincent Apps over the MCP protocol.
+
+It leverages the `mcp` utils in `@lit-protocol/vincent-sdk` to build a server from a Vincent App definition and then exposing it over the STDIO or HTTP transport.
+
+## Setup
+
+- Copy `vincent-app.example.json` to `vincent-app.json` or any other name you want and configure your Vincent App definition in it.
+- Copy `.env.example` to `.env` and fill in the values. Use absolute paths for the `VINCENT_APP_JSON_DEFINITION` value.
```
DashKash54, line 10: Do we need both can we simplify and just provide the env vars as keys in the json?

packages/mcp/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Running
+
+The recommended way to run the Vincent MCP server is using the npx commands below.
+
+But if you want to run it locally, you can build the package and run the server directly. Or run locally in development mode which will enable hot reloading, source code updates, etc.
```
DashKash54, line 16: Running locally with hot reloading is prob how most development will happen? We should add more details for this in the README?

packages/mcp/README.md
```diff
@@ -0,0 +1,92 @@
...
+```bash
+npx @lit-protocol/vincent-mcp http
+```
+
+In HTTP mode, the environment variables are configured on the server itself, not the client.
+
+These commands require the following environment variables to be set:
+- `VINCENT_APP_JSON_DEFINITION`: Path to your Vincent App definition JSON file
+- `PUBKEY_ROUTER_DATIL_CONTRACT`: The public key router Datil contract address
+- `VINCENT_DELEGATEE_PRIVATE_KEY`: The private key of the delegatee
```
DashKash54, line 39: We should add the definition of a delegatee here

packages/mcp/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "build:watch": "tsc --watch",
+    "dev:stdio": "tsx src/stdio.ts",
+    "dev:http": "tsx watch --env-file=.env src/http.ts",
+    "inspector": "npx @modelcontextprotocol/inspector",
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "mintRlI": "tsx --env-file=.env ./src/bin/mintRLINft.ts"
+  },
+  "dependencies": {
+    "@lit-protocol/constants": "^7.1.1",
+    "@lit-protocol/contracts-sdk": "^7.1.1",
```
DashKash54, line 38: We shouldn't be using the Lit JS SDK and replace it with the Vincent Contracts SDK @spacesailor24 
The idea of the Vincent SDK is to hide the complexity of the Lit network from the devs

packages/mcp/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@modelcontextprotocol/sdk": "^1.11.2",
+    "@t3-oss/env-core": "^0.13.4",
+    "ethers": "^5.8.0",
```
DashKash54, line 41: We need to standardize it are we using Viem in the Contracts SDK?

packages/mcp/src/bin/mintRLINft.ts
```diff
@@ -0,0 +1,101 @@
...
+const litContractClient = new LitContracts({
+  network: LIT_NETWORK.Datil,
+  signer: ethersSigner,
+});
+
+/**
+ * Represents information about a Capacity Credit. Includes the capacity token ID, requests per
+ * kilosecond, expiration details, and minting timestamp.
+ */
+interface CapacityCreditInfo {
```
DashKash54, line 24: Can we just sponsor all the dev transactions as well on Datil as it's free for us? And just use the relayer?
I wanna simplify the Devx as much as possible and we have a service which auto tops up the RLI NFT as well 
The less complexity the dev is exposed to the simpler for them and the better DevX, especially given that we already have this built out
This needs to be redone anyways for Naga since it involves real tokens

Proposal: When devs registers their Vincent Apps we add their Manager Address to the Relayer which allows us to use the PaymentDB for all their delegatees as well? @glitch003

packages/mcp/src/bootstrap.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * MCP Server Bootstrap for STDIO Transport
+ *
+ * This file is essential when using the stdio transport mechanism in the MCP server.
+ * It solves a critical problem: separating JSON-RPC communication from logs/debug output.
```
DashKash54, line 5: Doesn't the server handle the STDIO comms with the client by itself: https://modelcontextprotocol.io/docs/concepts/architecture#implementation-example
We need this explicit separation because of the additional logs/debug-- where are these coming from?

packages/mcp/src/bootstrap.ts
DashKash54, line 1: How does this work for remote servers which doesn't use STDIO to comm with the client-- how does the filtering happen for this scenario?

packages/mcp/src/env.ts
```diff
@@ -0,0 +1,30 @@
+import { createEnv } from '@t3-oss/env-core';
+import { z } from 'zod';
+
+// Ref: https://github.com/t3-oss/t3-env/pull/145
+const booleanStrings = ['true', 'false', true, false, '1', '0', 'yes', 'no', 'y', 'n', 'on', 'off'];
```
DashKash54, line 5: Patches Zod loophole?

DashKash54 (commented):
Left some preliminary comments

packages/mcp/.env.example
FedericoAmura: Vincent contracts use the pkp pubkey. The tool to get delegators fetches that public key and then uses the contracts sdk to get the token id (not needed) and the eth address (important). This last field is what users will tipically reason about unless they want to get deep into Lit

We can abstract it later in the vincent sdk, sure

packages/mcp/README.md
FedericoAmura: If we remove one, it should be the json as once we have the registry it will not be needed anymore, neither its env variable. However, putting the info it has in env variables is very annoying

And with the `npx` command you'll be able to run everything directly in the console with `export VINCENT_DELEGATEE_PRIVATE_KEY=0x123... npx @lit-protocol/vincent-mcp http`

In summary, the registry is going to replace the json automatically

packages/mcp/README.md
FedericoAmura: Details for local running are a few lines below

However, the ideal setup will be just running the `npx` command to start the mcp server. That way they don't even have to download this repo

packages/mcp/src/bootstrap.ts
FedericoAmura: This solves the problem that everything was fine using `console.log` until this STDIO transport came along. Whatever gets logged in the server to sdtio (for example, doing `console.log("I'm a log!")`) gets printed along the JSON-RPC messages that are passed from server to client. The protocol specification [explicitly forbids doing this](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#:~:text=The%20server%20MUST%20NOT%20write%20anything%20to%20its,stdin%20that%20is%20not%20a%20valid%20MCP%20message.) but everyone has been doing it since forever (even us in the JS SDK)

Some clients ignore this extra messages, others show errors to the user (claude desktop is here) and some break completely.
This is needed as a security net against any piece of code that wants to log something and the client cannot handle it gracefully

packages/mcp/src/env.ts
FedericoAmura: Exactly, makes it a easier to define `false` values on env variables. It was brought by @MaximusHaximus

packages/mcp/README.md
FedericoAmura: Added with a link to the doc where it explains it

packages/mcp/.env.example
FedericoAmura: Responded [here](https://github.com/LIT-Protocol/Vincent/pull/104#discussion_r2104283511)

packages/mcp/package.json
FedericoAmura: In v8, not in v7 which is the one that vincent uses

packages/sdk/typedoc.json
```diff
@@ -9,12 +9,13 @@
     "docs/Why-Vincent.md",
     "docs/Developers/Quick-Start.md",
     "docs/Developers/Custom-Tools.md",
+    "docs/Developers/AI-Integration.md",
```
DashKash54, line 12: Prob rename it to MCP since it's trendy and that what the page is as well?

packages/sdk/README.md
```diff
@@ -191,6 +191,89 @@ When JWT validation fails, descriptive error messages are thrown to help with tr
...
+import { mcp } from '@lit-protocol/vincent-sdk';
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+
+// Create a signer for your delegatee account
+const provider = new ethers.providers.JsonRpcProvider('https://mainnet.infura.io/v3/YOUR_INFURA_KEY');
+const delegateeSigner = new ethers.Wallet('YOUR_DELEGATEE_PRIVATE_KEY', provider);
+
+// Define your Vincent application with tools
+const vincentApp: mcp.VincentAppDef = {
+  id: '8462368',
```
DashKash54, line 217: This is the Vincent appId?

packages/sdk/README.md
```diff
@@ -191,6 +191,89 @@ When JWT validation fails, descriptive error messages are thrown to help with tr
...
+      name: 'checkBalance',
+      description: 'Check the balance of an account',
+      parameters: [
+        {
+          name: 'address',
+          type: 'address',
+          description: 'Ethereum address to check'
+        }
+      ]
+    }
```
DashKash54, line 247: The Tools will be eventually fetched from the Registry by just passing the appId? Or will we still allows custom tools which aren't part of the Vincent Registry for the appid?

packages/sdk/project.json
```diff
@@ -22,7 +22,7 @@
     "typedoc": {
       "executor": "nx:run-commands",
       "options": {
-        "commands": ["pnpm typedoc --watch", "node ./scripts/typedocServer.mjs"],
+        "commands": ["pnpm typedoc --skipErrorChecking", "node ./scripts/typedocServer.mjs"],
```
DashKash54, line 25: Why are we skipping the error checks?

packages/sdk/docs/Developers/AI-Integration.md
```diff
@@ -0,0 +1,111 @@
+---
+category: Developers
+title: AI Integration
+---
+
+# AI Integration
```
DashKash54, line 6: This page is specifically for MCP so we can change the title to reflect that?

packages/sdk/src/mcp/definitions.ts
```diff
@@ -0,0 +1,204 @@
...
+
+import { z, ZodRawShape } from 'zod';
+
+/**
+ * Supported parameter types for Vincent tool parameters
+ *
+ * These types define the valid parameter types that can be used in Vincent tool definitions.
+ * Each type has corresponding validation logic in the ZodSchemaMap.
+ */
+const ParameterType = [
```
DashKash54, line 19: We define valid param types using the Json schema as well in the Policy NPM package, why is it defined here separately?

packages/sdk/src/mcp/definitions.ts
DashKash54, line 1: @MaximusHaximus @spacesailor24 this seems to be related to the Registry can you guys please ensure the compatibility requirements

packages/sdk/src/mcp/server.ts
```diff
@@ -0,0 +1,133 @@
...
+  delegateeSigner: ethers.Signer,
+  vincentAppDefinition: VincentAppDef
+): McpServer {
+  const server = new McpServer({
+    name: vincentAppDefinition.name,
+    version: vincentAppDefinition.version,
+  });
+
+  Object.entries(vincentAppDefinition.tools).forEach(([toolIpfsCid, tool]) => {
+    server.tool(
```
DashKash54, line 124: Does this also include separately injected tools which aren't registered in the Registry for the appId?

packages/sdk/src/mcp/server.ts
DashKash54, line 1: We need to update it based on the ToolPolicySDK which has a pre-checker @MaximusHaximus @spacesailor24 ?

packages/mcp/src/stdio.ts
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env node
+/**
+ * Standard I/O server implementation for Vincent MCP
+ *
+ * This module provides a standard I/O (stdio) server implementation for the Model Context Protocol (MCP).
+ * It creates a server that communicates through standard input and output streams, making it suitable
+ * for integration with command-line tools, language models, and other processes that can communicate
+ * via stdio.
+ *
+ * The server loads a Vincent application definition from a JSON file and creates
```
DashKash54, line 10: **TODO:** Use the Registry after it's available

packages/mcp/src/extensions.ts
DashKash54, line 1: How have these extension functions been tested?

DashKash54 (commented):
Left some more comments, largely looks good!

packages/mcp/package.json
DashKash54: Also this isn't exposed to the devs it's only used by us?

packages/mcp/src/bin/mintRLINft.ts
glitch003: yeah we can do this - when they register a delagatee key in Lit we can add them to the payment DB, paid for by a master address that we put into the auto top up system.

packages/mcp/package.json
FedericoAmura: Exactly, when we migrate to v8 we can change it without breaking the public interface

packages/sdk/typedoc.json
FedericoAmura: It also has the OpenAI Agent Kit and was going to add the the connection to the claude and openai apis
So it is more than just MCP, hence that name

But can rename 👍🏻

packages/sdk/docs/Developers/AI-Integration.md
FedericoAmura: Same [here](https://github.com/LIT-Protocol/Vincent/pull/104#discussion_r2111419000)

PR #110, opened 2025-05-27
Feature - App, tool, and policy registry

glitch003:
App, tool, and policy endpoints implemented according to the spec, with accompanying tests.  

This PR may not be fully correct - we should review it against the spec, but it should be a good starting point for the registry to get us going.

### DIFF
--- a/packages/registry/.env.example
+++ b/packages/registry/.env.example
@@ -1,0 +1,2 @@
+PORT=3080
+MONGODB_URI=mongodb://localhost:27017/registry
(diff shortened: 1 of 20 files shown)


rsmets:
@glitch003, is the spec available to view publicly?

glitch003:
> @glitch003, is the spec available to view publicly?

Sure, it's published here: https://litprotocol.notion.site/Vincent-Off-chain-Service-Tech-Spec-1e36f449b04180ab8862e522ea03ee99

Appreciate any feedback or ideas!

PR #112, opened 2025-05-29
fix: temporary vincent tool and policy bundling fix

FedericoAmura:
SPENDING_LIMIT_POLICY_IPFS_ID=QmQCjBCemMAzVPZGAokwwxfT2cfa3jMYhX1YXkVYc1yed9
UNISWAP_SWAP_TOOL_IPFS_ID=QmSacbfpudFVjo6rsGKXhJU1cTywprmXwV3MNxjsmVhmsy


spacesailor24:
[This commit](https://github.com/LIT-Protocol/Vincent/pull/112/commits/19a4a5ff844e11576d3e2179c45529166c8ab57b) reverts the changes to the tool and policy `package.json`s. We need separate nx script for building the LA and the package. The flow of the nx scripts is: `action:build -> action:deploy -> build` as the `action` `build` and `deploy` scripts write the `src/generated` files that include the bundled LA code and the IPFS CID of the LA which needs to be included in the final `dist` build via `build`. We don't combine the `action:deploy` into `build` or make it `dependOn` because we don't want to force the dev to deploy to IPFS every time they build the package of run `pnpm build` at the project root

PR #115, opened 2025-06-02
Feature/drel 725 integrations with openai and anthropic responses apis

FedericoAmura:
# Description

This PR:
- Fixes session management (using TTL timestamps) removing the need for clients to call DELETE on them (some never did, some do at wrong times)
- Updates documentation on how to integrate with OpenAI and Anthropic responses APIs (were failing before due to incorrect session management or config)

### DIFF
--- a/packages/mcp/integrations/anthropic.ts
+++ b/packages/mcp/integrations/anthropic.ts
@@ -1,0 +1,91 @@
+import Anthropic from '@anthropic-ai/sdk';
+
+async function main() {
+  const anthropic = new Anthropic();
+
+  const msg = await anthropic.beta.messages.create({
+    betas: ['mcp-client-2025-04-04'],
+    model: 'claude-sonnet-4-20250514',
+    max_tokens: 1000,
+    mcp_servers: [
+      {
+        type: 'url',
+        url: process.env.VINCENT_MCP_URL, // The URL of the MCP server
+        name: 'vincent-app',
+        tool_configuration: {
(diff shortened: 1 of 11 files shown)


packages/mcp/integrations/anthropic.ts
```diff
@@ -0,0 +1,91 @@
+import Anthropic from '@anthropic-ai/sdk';
+
+async function main() {
+  const anthropic = new Anthropic();
+
+  const msg = await anthropic.beta.messages.create({
+    betas: ['mcp-client-2025-04-04'],
+    model: 'claude-sonnet-4-20250514',
```
Ansonhkg, line 8: should these be env vars?

packages/mcp/integrations/openAI.ts
```diff
@@ -0,0 +1,32 @@
+import OpenAI from 'openai';
+
+async function main() {
+  const client = new OpenAI();
+
+  const response = await client.responses.create({
+    model: 'gpt-4.1',
```
Ansonhkg, line 7: should this be env var?

Ansonhkg (commented):
do we have a doc to say that localhost doesn't work and we need to use ngrok to tunnel?

packages/mcp/integrations/openAI.ts
FedericoAmura: This is just for demo. Just to show how to integrate openai api with an mcp

PR #128, opened 2025-06-11
Feature/drel 740 optimize contract storage

DashKash54:
# What

1. Optimize PolicyVar setting by the user
2. Allow subset updates for permitted Apps
3. Remove `removeToolPolicyParameters()` since we can just reuse `setToolPolicyParameters()` to set the PolicyVars to zero

# Why

1. Remove duplicate logic and consolidate common steps for permitting App and updating PolicyVars
2. Since Apps can un-delete and enable their Apps anytime we should allows updating the PolicyVars irrespective of the App status
3. Removes one function and a bunch of duplicate tests

# Next

Will raise a few more PRs:
1. Comparing the gas of the new permit app v/s the old implementation. It might also include additonal gas optimizations
2. Optimization and functionality changes for the App Facet
3. Better Error messages and improved tests removing the redundancies

### DIFF
--- a/packages/vincent-contracts/src/VincentBase.sol
+++ b/packages/vincent-contracts/src/VincentBase.sol
@@ -12,6 +12,7 @@ contract VincentBase {
     error AppNotRegistered(uint256 appId);
     error AppVersionNotRegistered(uint256 appId, uint256 appVersion);
     error AppHasBeenDeleted(uint256 appId);
+    error AppVersionNotEnabled(uint256 appId, uint256 appVersion);
 
     /**
      * @notice Validates that an app exists
@@ -45,6 +46,14 @@ contract VincentBase {
         VincentAppStorage.AppStorage storage as_ = VincentAppStorage.appStorage();
         if (as_.appIdToApp[appId].isDeleted) {
             revert AppHasBeenDeleted(appId);
+        }
+        _;
+    }
+
+    modifier appEnabled(uint256 appId, uint256 appVersion) {
+        VincentAppStorage.AppStorage storage as_ = VincentAppStorage.appStorage();
+        if (!as_.appIdToApp[appId].appVersions[getAppVersionIndex(appVersion)].enabled) {
+            revert AppVersionNotEnabled(appId, appVersion);
         }
         _;
     }
--- a/packages/vincent-contracts/src/facets/VincentUserFacet.sol
+++ b/packages/vincent-contracts/src/facets/VincentUserFacet.sol
@@ -60,7 +60,11 @@ contract VincentUserFacet is VincentBase {
         string[] calldata toolIpfsCids,
         string[][] calldata policyIpfsCids,
         bytes[][] calldata policyParameterValues
-    ) external appNotDeleted(appId) onlyPkpOwner(pkpTokenId) onlyRegisteredAppVersion(appId, appVersion) {
+    ) external appNotDeleted(appId) onlyRegisteredAppVersion(appId, appVersion) appEnabled(appId, appVersion) onlyPkpOwner(pkpTokenId) {
+        if (toolIpfsCids.length == 0 || policyIpfsCids.length == 0 || policyParameterValues.length == 0) {
+            revert LibVincentUserFacet.InvalidInput();
+        }
+
         if (
             toolIpfsCids.length != policyIpfsCids.length || toolIpfsCids.length != policyParameterValues.length
         ) {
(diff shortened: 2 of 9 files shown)


packages/vincent-contracts/src/VincentBase.sol
```diff
@@ -49,6 +50,14 @@ contract VincentBase {
...
     }
 
+    modifier appEnabled(uint256 appId, uint256 appVersion) {
+        VincentAppStorage.AppStorage storage as_ = VincentAppStorage.appStorage();
+        if (!as_.appIdToApp[appId].appVersions[getAppVersionIndex(appVersion)].enabled) {
+            revert AppVersionNotEnabled(appId, appVersion);
+        }
+        _;
+    }
+
```
spacesailor24, line 60: Most of the functions that require this check are going to do additional reads from `VincentAppStorage.AppStorage`, and each time we read from `VincentAppStorage.AppStorage storage as_ = VincentAppStorage.appStorage();` it's `2,100` gas

This is the same for the `appNotDeleted` and `onlyRegisteredAppVersion` modifiers

DashKash54: Yes I was thinking the same, maybe we can combine these functions into a single modifier

packages/vincent-contracts/src/facets/VincentUserFacet.sol
```diff
@@ -82,44 +88,10 @@ contract VincentUserFacet is VincentBase {
...
 
-        // Verify that every tool provided is registered and every registered tool is provided
-        if (toolIpfsCids.length > 0) {
-            bytes32[] memory providedToolHashes = new bytes32[](toolIpfsCids.length);
-            for (uint256 i = 0; i < toolIpfsCids.length; i++) {
-                providedToolHashes[i] = keccak256(abi.encodePacked(toolIpfsCids[i]));
-                if (!newAppVersion.toolIpfsCidHashes.contains(providedToolHashes[i])) {
-                    revert LibVincentUserFacet.ToolNotRegisteredForAppVersion(appId, appVersion, toolIpfsCids[i]);
-                }
-            }
```
spacesailor24, line 104: This was validating that every tool provided by the user to `permitAppVersion` is specified as a tool for the App version they're permitting. Without this, a User could provide any tools as long as the length matches the expected amount of tools for the App version

packages/vincent-contracts/src/VincentBase.sol
spacesailor24: I'm not sure the overlap between all the functions and modifiers, but if it's not great enough, the cheapest option would be inlining all the checks into each function, sacrificing readability

DashKash54: To be addressed in "Next Steps" PR point 1

packages/vincent-contracts/src/facets/VincentUserFacet.sol
```diff
@@ -434,17 +293,25 @@ contract VincentUserFacet is VincentBase {
...
+                    if (keccak256(abi.encodePacked(policyIpfsCids[i][k])) == hashedToolPolicy) {
+                        revert LibVincentUserFacet.DuplicateToolPolicyIpfsCid(appId, appVersion, toolIpfsCid, policyIpfsCids[i][k]);
+                    }
+                }
+
                 // Step 5: Store the policy parameter metadata
                 toolPolicyParameterValues[hashedToolPolicy] = policyParameterValues[i][j];
 
                 // Step 5.1: Emit an event for tracking
+                // TODO!: Add Policy IpfsCid to the event
```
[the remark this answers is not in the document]
DashKash54: To be addressed in "Next Steps" PR point 3

packages/vincent-contracts/src/VincentBase.sol
DashKash54: Subsequent reads will be warm thus significantly cheaper. If we do this it will be at the cost of readability and code duplication so we should keep it. Additionally the cost of these warm read is dwarfed by the cost of storage for writing policyVars. So in the best case it will save a couple percentages of gas per permit. The real cost saver would be not storing the entire policyVar CBOR on chain but rather on IPFS and just storing the hash on chain. But this will add a whole new storage layer, IPFS so we shouldn't do this pre-mature optimization unless we have a lot of users and the chain becomes a bottleneck for scaling Vincent.

DashKash54:
Subsequent reads will be warm thus significantly cheaper. If we do this it will be at the cost of readability and code duplication so we should keep it. Additionally the cost of these warm read is dwarfed by the cost of storage for writing policyVars. So in the best case it will save a couple percentages of gas per permit. The real cost saver would be not storing the entire policyVar CBOR on chain but rather on IPFS and just storing the hash on chain. But this will add a whole new storage layer, IPFS so we shouldn't do this pre-mature optimization unless we have a lot of users and the chain becomes a bottleneck for scaling Vincent.

DashKash54:
The other major chunk of the permitAppVersion gas cost comes from storage:
1. newAppVersion.delegatedAgentPkps.add(pkpTokenId)
2. agentStorage.permittedApps.add(appId)
3. us_.userAddressToRegisteredAgentPkps[msg.sender].add(pkpTokenId)
~70k each all three have to be Enumberable Set instead of arrays since we need to check for duplicates and removal in the first case

packages/vincent-contracts/src/facets/VincentUserFacet.sol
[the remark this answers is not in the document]
DashKash54: Addressed in this PR: https://github.com/LIT-Protocol/Vincent/pull/188

PR #131, opened 2025-06-12
SPIKE: Lit action esbuild configuration to `vincent-tool-sdk`

MaximusHaximus:
Initial spike of pulling esbuild configuratio n + plugins into a sane, non self-executing function scope in `vincent-tool-sdk

cc: @Ansonhkg


packages/vincent-tools/vincent-tool-sdk/src/lit-action-builder/plugins/polyfillNode.ts
```diff
@@ -0,0 +1,67 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+import { polyfillNode } from 'esbuild-plugin-polyfill-node';
+
+export const basePluginConfig = {
```
MaximusHaximus, line 5: Our current esbuild-plugin-polyfill-node configuration is wrong - it uses incorrectly cased strings an leaves a ton of stuff enabled that must be explicitly disabled

PR #138, opened 2025-06-16
Vincent Tool Client v2

MaximusHaximus:
# Description

### Implements Vincent Tool Client v2
- Takes a bundledVincentTool object and an ethers signer as an argument, and returns a strongly typed client object that handles execute() and precheck() logic.

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] Updated e2e tests to use the new client and ran them
- [X] Added a precheck() test case to the e2e tests

# Checklist:


packages/vincent-tools/vincent-tool-erc20-approval/src/generated/vincent-bundled-tool.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD.
```
FedericoAmura, line 2: nit: This is not dynamic anymore. Can it be removed from the esbuild generation? Maybe place this logic in/near the index that exports it

FedericoAmura (approved):
Left a few minor comments. Let's ship this

packages/vincent-tools/vincent-tool-erc20-approval/src/generated/vincent-bundled-tool.ts
MaximusHaximus: There are two goals with this:
1. Devs don't need to know about `asBundled....()` -- if they don't use that method, their tool/policy will not be well-formed, so by generating it using tooling we are codifying that requirement by the bundling step instead of purely by convention.
2. The file only exists when there is metadata file for it to reference; since this imported from that metadata file, and that file is generated, I thought it made the most sense to have them both be generated at the same time.

The intention is to move the `lit-action.ts` file to be generated also, actually -- that way developers won't need to know about `vincentPolicyHandler()` or `vincenToolHandler()` either.

So the generated code will be:
- lit-action.ts w/ the `handler` wrapper function
- vincent-bundled-<X> w/ the bundled tool/policy including ipfs cid
- metadata.json w/ the ipfsCid (and for policies, we will also include the uiSchema/jsonSchema keys there

The only file that the tool/policy developer will define is the actual vincent-tool.ts/vincent-policy.ts file that uses `createVincentTool`, `createVincentToolPolicy`, or `createVincentPolicy()`, and their uiSchema/jsonSchema files.

PR #165, opened 2025-06-19
Remove prefix api/v1 from registry routes

MaximusHaximus:
# Description
This fixes a regression that happened when moving the package to the new repo structure. Routes in OpenAPI docs are no longer prefixed with api/v1, which fixes the 'openAPI playground'.


awisniew207 (approved):
:saluting_face:

PR #193, opened 2025-07-02
MCP doc fixes

FedericoAmura:
# Description

This PR fixes a few things in MCP docs, removes some unncessary non-null assertions and moves integration docs to MCP public document instead of being actual code (thus removing their dependencies)

## Type of change

- [x] This change is a documentation update

# How Has This Been Tested?

- [x] Opening the links
- [x] Checking docs
- [x] Using the MCP with the uniswap swap app

# Checklist:


FedericoAmura:
> 👍 🚢 , but I'm not too keen on inlining the entire examples into the MCP.md doc.
> 
> IMO, moving forward:
> 
> * We should add typedoc support to the `mcp-server` project so it gets it's own API Docs category
> * We should keep the code examples as dev dependencies that are in a directory that is typechecked but not part of the build output, so we know that they are valid and copy/pasting them somewhere won't result in a type error.  Incidentally, I copy/pasted them into my own project and immediately got a TS error about possibly undefined values, so this isn't just a theoretical concern.
> * We could emit the example code directly in the `mcp-server` typedoc output as module exports, and _link to them_ from the main MCP.md file instead of in-lining them there.

Noted. Will try to move now in the MCP PR

PR #198, opened 2025-07-06
Feat: Simplify Contract-SDK

DashKash54:
# What

1. Nuke existing Contracts-SDK
2. New Contracts-SDK should be minimalistic- just a Viem client that exposes a signer and provides a single interface for all the Vincent contracts functions
3. Fix the broken contracts-sdk package
4. Add E2E tests for the Contracts-SDK

# Tests

Since the contracts have already been tested in the Solidity tests, especially the edge cases, so we're just adding a single test file which use the new Contract-SDK to:
1. Register an App
2. Modify the delegatee
3. Publish a new App version
5. Permit an App
6. Update PolicyVars
7. Upgrade to the new App version
8. Revoke App


MaximusHaximus (changes requested):
:+1: Looking good -- I added just one critical thing in-line, regarding gas estimation.

Regarding open API implementation questions:
> A single unified class-based VincentContracts approach will contains over 20 functions which won't be tree-shakable while managing only a single state Ethers Contract which is initialized with the provided Signer (EOA, PKPEthersWallet). Would be better to proceed with a functional design and the user needs to pass the Signer (EOA, PKPEthersWallet) on every instance and we internally initialize the Ethers Contract with it?

I agree -- we want this SDK to be stateless if at all possible; don't make a class that wraps all of the functions IMO -- but I don't think tree-shaking is really the driving concern here, since the ABI itself isn't tree-shakeable anyway and there's only _one contract involved_.  There are other good reasons to keep this module defined as a functional style though:
- It is trivial for someone to wrap calls to our base functions with some simple code that always provides the `signer` in their user-space code.
- It allows for powerful composition and eliminating coupling for consumers of the SDK
  - if instances of the SDK must be created, and they are stateful (signer-specific), then consumers will have to create _and maintain_ references instances of the contract SDK that will just be shadows instances of their signers.
  - If someone is using multiple signers and the SDK is stateless, they _just_ need to call our SDK methods w/ the appropriate signer instead of needing to maintain a mapping of the correct contract-sdk instance per user context.

> Most of the functions are just doing the same steps: manual gas estimation, calling the smart contract function, parsing logs for write functions to return the value (see registerApp()), decoding the error. Could we make a clever use of generics or use bind?

- I don't think we've got enough repetition here to really justify a composition layer for these functions -- we have a finite set of them and the inputs and outputs are entirely known. Let's just enumerate them explicitly and keep everything straightforward type-system-wise.
- Gas estimation should be done using a simple wrapper function, though -- see my in-line comment about us needing to do that anyway to always add 10% to the estimated amount.
- Given that the log/events output structure is going to be different for each function, I don't see any way around making their parsing explicit for each function that doesn't involve a rube-goldberg system that wraps basic string parsing. We may as well just handle the logs in each wrapper function IMO.




> Should we separate the User and App functions for less boilerplate? Basically maintaining a standard SDK design for the Vincent monorepo
IMO, I think that the method names are fairly self explanatory.  Let's land this as a flat set of methods for now -- if we want to wrap them into something with more explicit namespaces, we can always do that as a non-breaking addition to the SDK in a patch release.  We will likely find that there are few enough actions, and those actions scopes are so obviously coupled to the role that is calling them, that grouping the few functions we have together like this is not useful and just adds an extra layer of complexity to interacting with the SDK for consumers.

packages/libs/contracts-sdk/src/facets/User.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * Sets tool policy parameters for a specific app version
+ * @param signer - The ethers signer to use for the transaction. Could be a standard Ethers Signer or a PKPEthersWallet
+ * @param args - Object containing pkpTokenId, appId, appVersion, toolIpfsCids, policyIpfsCids, and policyParameterValues
+ * @param overrides - Optional override params for the transaction call like manual gas limit
+ * @returns The transaction hash and a success flag
+ */
+export async function setToolPolicyParameters({
```
MaximusHaximus, line 113: 🤔 Since CBOR2 encoding of policyParameters is an internal implementation detail of the data layer, we want consumers of the SDK to pass in plain javascript objects for their policy parameters -- which we CBOR2.encode() and then pass along to the contracts -- and we should assume that any policy parameters that we fetch are CBOR2 encoded -- so we can CBOR2.decode() them and return the resulting plain JS objects to consumers who are fetching them.  If anything _other_ than CBOR2 data is written to the contracts, it won't be usable by our policy engine running on the VincentToolClient and our Lit Action wrappers -- so the encoding/decoding should be as close to the contracts as we can make it, API wise.

We should put CBOR2 encoding  and decoding inside these methods, so that users of the contracts SDK can pass plain javascript objects to these methods, and get plain javascript objects back when they fetch policy parameters

packages/libs/contracts-sdk/src/facets/UserView.ts
```diff
@@ -0,0 +1,111 @@
...
+    const pkpTokenId = utils.parseUnits(args.pkpTokenId, 0);
+    const appId = utils.parseUnits(args.appId, 0);
+
+    const tools = await contract.getAllToolsAndPoliciesForApp(pkpTokenId, appId);
+
+    return tools.map((tool: any) => ({
+      toolIpfsCid: tool.toolIpfsCid,
+      policies: tool.policies.map((policy: any) => ({
+        policyIpfsCid: policy.policyIpfsCid,
+        policyParameterValues: policy.policyParameterValues,
```
MaximusHaximus, line 104: This map should return `policyParameterValues: CBOR2.decode(policy.policyParameterValues)` per my other comment :+1:

MaximusHaximus (changes requested):
Nice work -- this is coming together nicely 💖 

I added some comments re:  handling CBOR2 encoding/decoding of policy parameters in-line.

Also, consumers of our APIs don't typically manage their delegatees using pkp token IDs -- instead, they usually identify the PKP using their eth address.  So far we don't expose PKP token IDs anywhere else to direct API consumers, and if we force consumers of these methods to use token IDs, they will need to figure out that they need to import the main LIT JS contracts SDK just to transform the PKP eth Address to PKP Token Id in most cases.  I think we should take `pkpEthAddress` as the argument and look up the tokenId inside these methods.

packages/libs/contracts-sdk/package.json
```diff
@@ -15,11 +15,23 @@
...
-    "pino-pretty": "^13.0.0"
+    "@types/jest": "^29.5.12",
+    "jest": "^29.5.0",
+    "pino-pretty": "^13.0.0",
+    "ts-jest": "^29.1.0"
   },
   "dependencies": {
-    "@lit-protocol/logger": "custom-logger"
+    "@lit-protocol/logger": "custom-logger",
+    "@lit-protocol/pkp-ethers": "^7.2.0",
```
MaximusHaximus, line 34: These should be dev deps if they are only relevant to testing :+1: -- this is why NX is throwing a Lint error about them in the CI check; if they aren't used by any code inside the `build` output directory, then they are considered incorrectly placed to be in standard `dependencies` :+1:

@Ansonhkg What is the `custom-logger` tag of `@lit-protocol/logger` and why do we have it just for this package? 🤔

Ansonhkg:
I don’t necessarily agree that this simplifies anything. The previous createChainClient function served as a composer for various already-functional utilities. Each of these functions takes in a Viem accountOrWalletClient object, which can easily be swapped out for Ethers wallet/signer. They also accept a network context, including the ABI, chain config, and rpc url.

However, this PR seems to keep us moving forward, i'm also not against it.

MaximusHaximus (approved):
:+1:

packages/libs/contracts-sdk/package.json
MaximusHaximus: :+1: `pkp-ethers` is only used as a dev dependency right now since we get the signer passed into the methods, so we can :hocho: the `custom-logger` dependency and move `@lit-protocol/pkp-ethers` to `devDependencies` @DashKash54

MaximusHaximus (approved):
:+1: 🚢

PR #212, opened 2025-07-13
Feat: Follow up contracts sdk simplification

DashKash54:
# Description

Add more utility functions and API simplifications to the new Contracts-SDK

1. Utility `getAppIdByDelegatee`
2. Graceful empty return value handling instead of throwing an error to differentiate between actual contract errors and empty return values

# Tests

1. New test for `getAppIdByDelegatee


MaximusHaximus:
Closed - not doing

PR #213, opened 2025-07-14
Feat: Contract utils functions + Diamond update

DashKash54:
# Description

Add support for utility contract functions:
1. Pagination update to only accept the `offset` as a param and use the internal `PAGE_SIZE` as the preset page limit for the following functions:
    1. `getDelegatedAgentPkpTokenIds`
    2. `getAppsByManager`
    3. `getAllRegisteredAgentPkps`
    4. `getAllPermittedAppIdsForPkp`
2. Simplified contract functions to follow the separation of concerns principle also makes the Contracts-SDK API simpler:
    1. `getAppVersion` no longer returns `delegatedAgentPkpTokenIds` since it can be a huge array
    2. `getAppsByManager` no longer return app details since it can be found from `getAppById`
    3. Added new function `setDelegatee` to remove & add an array of delegatees at once thus reducing the number of tx calls. This also allows to remove all delegatees from an app at once
3. Diamond Update Script- replaced with a SmartUpdate script that automatically determines the function selectors to be added, removed or replaced by comparing the current undeployed contracts and the old deployed contracts
4. Simplified the upgrade deployment script and removed redundancies:
    1. Deploying only to Datil and removing all the dev, test deployment logic.
    2. Remove unused post-build scripts
5. Generated new ABIs for the deployed facets

Use the fixed contract update scripts for multiple deployments various facets.

# Tests

1. Added contracts & Jest tests for the above point 1

# Others

See this [comment](https://github.com/LIT-Protocol/Vincent/pull/213#issuecomment-3146875841) for latest changes after the main merge


packages/libs/contracts-sdk/postbuild-scripts/00-abi-conversions.ts
spacesailor24, line 1: Just curious why we are removing this? I thought the contracts SDK wrapper required these outputs?

DashKash54: It doesn't use it anymore, I replaced it already and merged to main

packages/libs/contracts-sdk/contracts/facets/VincentAppFacet.sol
```diff
@@ -164,6 +164,50 @@ contract VincentAppFacet is VincentBase {
...
+        VincentAppStorage.App storage app = as_.appIdToApp[appId];
+
+        address[] memory currentDelegatees = app.delegatees.values();
+        
+        // Remove all current delegatees
+        for (uint256 i = 0; i < currentDelegatees.length; i++) {
+            app.delegatees.remove(currentDelegatees[i]);
+            as_.delegateeAddressToAppId[currentDelegatees[i]] = 0;
+            emit LibVincentAppFacet.DelegateeRemoved(appId, currentDelegatees[i]);
+        }
```
spacesailor24, line 190: I think storing `currentDelegatees[i]` in a local var as done in the next for loop offers some gas savings

packages/libs/contracts-sdk/test/VincentContracts.spec.ts
```diff
@@ -325,16 +321,17 @@ describe('VincentContracts', () => {
...
+    // const agentPkpsResult = await getAllRegisteredAgentPkps({
+    //   signer: userSigner,
+    //   args: {
+    //     userAddress: process.env.TEST_USER_PKP_ADDRESS!,
+    //     offset: '0',
+    //   },
+    // });
+    // console.log('Agent pkps result:', agentPkpsResult);
+    // expect(agentPkpsResult.length).toBeGreaterThan(0);
+    // expect(agentPkpsResult[0]).toBe(process.env.TEST_USER_AGENT_PKP_TOKEN_ID!);
```
spacesailor24, line 334: Commented out code

packages/libs/contracts-sdk/contracts/facets/VincentUserViewFacet.sol
```diff
@@ -134,40 +153,52 @@ contract VincentUserViewFacet is VincentBase {
...
+        if (end > nonDeletedCount) {
+            end = nonDeletedCount;
+        }
+
+        uint256 resultCount = end - offset;
+        uint256[] memory result = new uint256[](resultCount);
+
+        for (uint256 i = offset; i < end; i++) {
+            result[i - offset] = permittedAppSet.at(i);
         }
```
spacesailor24, line 199: I don't think this is properly accounting for the deleted Apps? It looks like if there were 3 permitted Apps and App 2 was deleted, then the function would return App 1 and 2. Because `nonDeletedCount` would be `2` and the for loop would return `permittedAppSet.at(1);` and `permittedAppSet.at(2);`

I didn't see a test in `VincentUserFacet.t.sol` that calls this function with deleted Apps

DashKash54: Nice catch, fixed and added a test: `testPermittedApps_AppHasBeenDeleted`

packages/libs/contracts-sdk/script/UpdateFacet.sol
```diff
@@ -7,235 +7,299 @@ import {console2} from "forge-std/console2.sol";
...
+            testSelector = VincentAppFacet.registerApp.selector;
+        } else if (compareStrings(facetName, "VincentAppViewFacet")) {
+            testSelector = VincentAppViewFacet.getAppById.selector;
+        } else if (compareStrings(facetName, "VincentUserFacet")) {
+            testSelector = VincentUserFacet.permitAppVersion.selector;
+        } else if (compareStrings(facetName, "VincentUserViewFacet")) {
+            testSelector = VincentUserViewFacet.getPermittedAppVersionForPkp.selector;
+        } else {
+            return false;
+        }
```
spacesailor24, line 232: This relies on the identifying function’s signature never changing. If the function is renamed, its parameters are changed, or it’s removed entirely, the script won’t be able to identify the facet

A simple fix might be just adding a facet name property onto all the facets, something like:
```
function facetName() external pure returns (string memory);
```

DashKash54:
New changes after the merge:
1. Remove `pageOpts` instead accepting `offset` directly since we removed limits. Having a nested object for a single param seems an overkill @MaximusHaximus 
2. Remove `AppWithVersions` since we're only returning appId and count and then you can make a second call to fetch the version details @a1ttech @MaximusHaximus 
3. New function `getAppIdByDelegatee` @a1ttech 
4. [Potential TODO] Accept offset in the Dashboard @a1ttech? https://github.com/LIT-Protocol/Vincent/pull/213/files#diff-6044cf5f0377cf5031fec27314d56cb164a50d15e63575398a2bd247480008bd
5. [Potential TODO] Accept offset in the MCP server @FedericoAmura? https://github.com/LIT-Protocol/Vincent/pull/213/files#diff-1d64d4fbe8e7267751dd5366b9f0be8bc37139332dd495f8912a730fc38d25ee

PR #217, opened 2025-07-17
fix(contracts-sdk): Fix incorrect function signature for `encodePermssionDataForChain()`

MaximusHaximus:
Bugfix!


awisniew207 (approved):
let's do it

PR #221, opened 2025-07-18
feat: Tool compatibility version checks

MaximusHaximus:
# Description
### Adds version compatibility checks for tools and policies, both at composition time (when binding policies to tools) and when executing tools using the VincentToolClient.
- Added a VINCENT_TOOL_API_VERSION semver constant to the tool-sdk, which is added to `BundledVincentTool` and `BundledVincentPolicy`
- Added a semver `major` comparison assertion so that when tools, policies, and the VincentToolClient are interacting, we throw an error if the various components don't at least have a `major` level semver match
- VincentToolClient throws an error if the tool you are using with it doesn't match its major semver. It provides its semver to the tool's LA runtime via jsParams when it calls it
- vincentToolHandler LA wrapper throws if the toolApiVersion isn't the same `major` semver. It passes the semver to the policies when it evaluates them
- vincentPolicyHandler LA wrapper throws if the toolApiVersion isn't the same `major` semver. This is unlikely to happen since the `createVincentTool()` method should've thrown an error when someone tried to compose the policy with a different major semver than the tool -- but...Just In Case :tm:

This system replaces the typescript-only hidden 'unique Symbol branding' that we were putting on BundledVincentPolicy and BundledVincentTool with much stronger runtime guarantees.
This also solves the issue of Symbol() comparisons not matching even if the tool-sdk version a tool was authored with was just slightly different than the tool-sdk version your current VincentToolClient is using. Now it's semver-major restricted based on the tool API which doesn't necessarily match the package semvers directly.

#### Imports and Exports hygiene
- Added strict eslint rules that ensure that types are always imported and exported using the `type` keyword -- which is critically important for a few different reasons in a modern TS repo using `isolatedModules`, which is necessary for `composite` configurations.
- Enabled basic imports sorting in the strict eslint rule config
- Updated `app-sdk`, `tool-sdk`, `contracts-sdk`, `registry-sdk`, `registry-backend` to use the strict eslint rules and ran --fix on all of them
- 
#### Base64 / Uint8Array handling in app-sdk
- Our JWT implementation was Buffer dependent and manually implemented some things that are part of `ethers/utils` -- which we already depend on and use in the same files!
- Replaced the bespoke Buffer-dependent code with code that uses Buffer if it's available, otherwise uses `atob`, so no Buffer polyfill is needed for the jwt code (which runs in the browser env!)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] Ran all e2e tests
- [X] Verified all type inference tests are passing

# Checklist:

- [X] I created a [release plan](https://nx.dev/recipes/nx-release/file-based-versioning-version-plans) (`nx release plan`) describing my changes and the version bump
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


eslint.config.strict.js
```diff
@@ -48,7 +48,7 @@ module.exports = [
       'import-x/no-duplicates': ['error'],
       'import-x/no-extraneous-dependencies': ['off'], // NX provides already
       'import-x/no-relative-packages': ['error'],
-      'import-x/no-unresolved': ['error'],
+      'import-x/no-unresolved': ['off'], // Fails to resolve `workspace:*` only in GH actions :(
```
FedericoAmura, line 51: 😢

packages/libs/tool-sdk/src/lib/assertSupportedToolVersion.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function assertSupportedToolVersion(
+  toolVersionSemver: string | undefined,
+): asserts toolVersionSemver is string {
+  if (!toolVersionSemver) {
+    throw new Error('Tool version is required');
+  }
+
+  // Check if the tool's API version has the same major version as the current API version
+  if (major(toolVersionSemver) !== major(VINCENT_TOOL_API_VERSION)) {
```
FedericoAmura, line 13: I think the check should also include >= for minor and patch when necessary

The tool might require a new feature added at a certain minor or a bug fix, which are not major changes

packages/libs/tool-sdk/src/lib/constants.ts
```diff
@@ -0,0 +1 @@
+export const VINCENT_TOOL_API_VERSION = '2.0.0' as const;
```
FedericoAmura, line 1: Shouldn't we get this from its package.json? We are going to forget updating this on release or version bump

packages/libs/tool-sdk/src/type-inference-verification/allow-deny-test-cases-tool.ts
```diff
@@ -658,7 +659,7 @@ export const testPrecheckAsyncWithoutAwait = () => {
     // @ts-expect-error - Missing return from async function that calls other async functions
     precheck: async (params, { succeed }) => {
       // Call async function without await or using its result
-      fetchData();
+      await fetchData();
```
FedericoAmura, line 662: Outdated comment

packages/apps/toolpolicies-e2e/project.json
```diff
@@ -9,12 +9,16 @@
...
     },
-    "test-e2e": {
+    "test-e2e-deploy-toolPolicies-in-other-projects-and-build-dist": {
       "dependsOn": [
         "policy-spending-limit:action:deploy",
         "tool-uniswap-swap:action:deploy",
-        "tool-erc20-approval:action:deploy"
-      ],
+        "tool-erc20-approval:action:deploy",
+        "^build"
```
glitch003, line 17: nice to see this fix so we don't have to manually rebuild

packages/libs/tool-sdk/src/lib/constants.ts
glitch003: might be good to put this into the docs somewhere.  like make it clear what the different versions are in the diff packages and how they interact.  like "major version of the tool-sdk is compatible with app-sdk" etc, all versions right now are compatible with VINCENT_TOOL_API_VERSION 2.0.0, etc.

glitch003 (approved):

PR #227, opened 2025-07-21
user docs edits

a1ttech:
introduced an "intro" page for the user docs and simplified the "key concepts" section


MaximusHaximus (approved):
:+1: Rock on, thanks G!

PR #264, opened 2025-08-05
feat: Update to latest JS SDK to fix walletconnect

glitch003:
Simple PR - I released v7.2.3 of the JS SDK which fixes walletconnect with EIP-1559 txns.  This PR just updates the dependencies to use that package.


packages/libs/app-sdk/package.json
```diff
@@ -68,7 +68,7 @@
     "*.md"
   ],
   "devDependencies": {
-    "@lit-protocol/pkp-ethers": "^7.2.0",
+    "@lit-protocol/pkp-ethers": "^7.2.3",
     "@lit-protocol/types": "^7.0.8",
```
FedericoAmura, line 72: Can we update types in here too?

FedericoAmura (approved):

packages/libs/app-sdk/package.json
glitch003: sure, i updated the PR to do it everywhere including here

PR #281, opened 2025-08-14
feat: add delegatee addresses to the payment DB

glitch003:
Edit: I added a new endpoint and am calling it after publish.  We also need to call it when they add delegatees.

The goal is to make it so that app devs don't think about payment for using Lit, or RLI NFTs.

This PR is a start to that goal.  It uses the existing relayer to add delegatee addresses to the payment DB, where our "master address" will pay for all delegatees.  We will add this "master address" to the lit-scheduled-tasks / auto-top-up repo so that it's always funded with a current RLI NFT.

I'm not sure where exactly to put this but I made my best guess - we can do it whenever the app gets created or edited.  Another alternative is to only do it when the app is published (like, add a new endpoint for this purpose) and call it from the frontend after publish succeeds.  

I don't think we really need to much auth here.  We can start with basically none just to get this working, and then add a simple check that hits the chain and gets the delegatee addresses from there instead of the user supplied ones.  It's debatable if this actually needs any auth at all right now - anyone can create a vincent app with any delegatee addresses.

### DIFF
--- a/packages/apps/registry-backend/package.json
+++ b/packages/apps/registry-backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lit-protocol/registry-backend",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "private": true,
   "scripts": {
     "prebuild": "node scripts/copy-api-html.js",
@@ -21,9 +21,9 @@
   },
   "dependencies": {
     "@lit-protocol/flags": "^2.1.0",
-    "@lit-protocol/vincent-app-sdk": "workspace:*",
-    "@lit-protocol/vincent-contracts-sdk": "workspace:*",
-    "@lit-protocol/vincent-registry-sdk": "workspace:*",
+    "@lit-protocol/vincent-app-sdk": "^2.0.0",
+    "@lit-protocol/vincent-contracts-sdk": "^1.0.1",
+    "@lit-protocol/vincent-registry-sdk": "^4.1.0",
     "@t3-oss/env-core": "^0.13.6",
     "cors": "^2.8.5",
     "debug": "^4.4.1",
--- a/packages/libs/registry-sdk/package.json
+++ b/packages/libs/registry-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lit-protocol/vincent-registry-sdk",
-  "version": "4.0.0",
+  "version": "4.1.0",
   "description": "Vincent Registry API definitions and generated code",
   "packageManager": "pnpm@10.7.0",
   "private": false,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -240,7 +240,7 @@ importers:
         version: 7.2.3(ethers@5.8.0(bufferutil@4.0.9)(utf-8-validate@5.0.10))
       '@lit-protocol/lit-node-client':
         specifier: ^7.2.3
-        version: 7.2.3(@walletconnect/modal@2.7.0(@types/react@19.1.8)(react@19.1.0))(bufferutil@4.0.9)(ethers@5.8.0(bufferutil@4.0.9)(utf-8-validate@5.0.10))(utf-8-validate@5.0.10)
+        version: 7.2.3(@walletconnect/modal@2.7.0)(bufferutil@4.0.9)(ethers@5.8.0(bufferutil@4.0.9)(utf-8-validate@5.0.10))(utf-8-validate@5.0.10)
       viem:
         specifier: 2.29.2
         version: 2.29.2(bufferutil@4.0.9)(typescript@5.8.3)(utf-8-validate@5.0.10)(zod@3.25.64)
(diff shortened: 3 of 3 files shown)


packages/apps/registry-backend/src/lib/express/app/routes.ts
```diff
@@ -20,6 +20,23 @@ function generateRandomAppId(): number {
   return Math.floor(Math.random() * (10_000_000_000 - 1000)) + 1000;
 }
 
+async function addDelegateesToPaymentDB(delegateeAddresses: string[]): Promise<void> {
+  // hit the relayer server to add the delegatee addresses
+  const headers = {
+    'api-key': process.env.LIT_RELAYER_API_KEY!,
```
glitch003, line 26: I know using the ! is not correct, but how can I set it up so that like, the server won't even start if these env vars are not present?  and then I don't need to check for their presence every time I use them?

FedericoAmura: Exactly, `createEnv` in that file does exactly what you want @glitch003 . It can also apply validations such as format, optional, default value and make them the right type

Then just read the env var from that file, [for example here](https://github.com/LIT-Protocol/Vincent/blob/main/packages/apps/registry-backend/src/lib/express/index.ts#L17)

MaximusHaximus:
Per your message, I think that rather than trying to embed this into our existing routes, we should define a new route (or routes, if we want to remove the payment entry when someone removes a delegatee from an app) that provides the pass-through for calling the add-users relayer endpoint with our api-key.  This is a lot more flexible and makes dealing with failure modes a lot easier from the frontend.  It also keeps our API endpoints as pure 'I edit Registry entries' endpoints, and those edits then get 'published' to on-chain, which we have worked hard to enforce across the system for sanity's sake.

If we make this an explicit route, then the frontend knows that creating or editing an app document is one atomic operation, and adding it to the payment contract is another one. There will be a clear route for how to handle errors adding the payment contract entry.  There are some pit-of-sadness failure modes that happen if we put this logic into the existing endpoints, since the call to the relayer can fail even though edits have been made successfully, or an app has been created successfully on the registry.  For example, if creating a new app succeeds, but adding the delegatee fails and it is a dedicated endpoint, then the frontend can retry the _just the operation to add the payment entry_ a few times, or provide an alert to the consumer that the app was edited or created successfully, but the delegatee is not yet useable because it hasn't been 'funded' for usage of the network.  It also makes it quite simple to do it during publishing of changes to delegatees to on-chain; an app owner may change delegatee addresses many times on the off-chain doc but _never publish any of those changes to the chain_ -- in that case, we don't want to have added all of those addresses to the payment contract even though they were never actually delegatees -- we get a lot of benefits from avoiding tying the editing of the registry docs to the management of the payment contract IMO.

My proposal is:
1. Define new endpoint(s) for adding (and removing) payment entries for delegatees to the REST API; these will have no implied document editing capabilities, but we can do the most basic checks that already exist to ensure that the endpoints are still coupled to the target app existing and having the delegatee address on the doc (for adding) -- and you must own the app to use them.
2. Add logic to the existing frontend flows that shadow our existing code that updates delegatees _on-chain entries_ -- so that we also modify their entries in the payment contract _when we go to write them on chain_.
3. If someone removes a delegatee from their app, we can make a request from the frontend to remove the payment entry to avoid keeping 'ghost' delegatees in the master payment list forever which would be a nice-to-have that would be easy to implement given there are dedicated endpoints for it.

packages/apps/registry-backend/.env.example
```diff
@@ -4,3 +4,5 @@ MONGO_DB_NAME=testing
 IS_DEVELOPMENT=true
 CORS_ALLOWED_DOMAIN=http://localhost:3000
 DEBUG=express:*
+EXPECTED_AUDIENCE=localhost:3000
```
glitch003, line 7: when i started the registry backend it complained that these things were missing so i added them to the .env.example

packages/apps/registry-backend/.env.example
```diff
@@ -1,6 +1,8 @@
 PORT=3000
-MONGODB_URI=mongodb://localhost:27017/registry
+MONGODB_URI=mongodb://localhost:27017/registry?replicaSet=rs0
```
glitch003, line 2: mongo wouldn't let me save anything until i turned on replica sets by adding this to `/opt/homebrew/etc/mongod.conf`

```
replication:
  oplogSizeMB: 2000
  replSetName: rs0
```

and then running `rs.initiate()` in the mongosh REPL

packages/apps/app-dashboard/src/config/env.ts
```diff
@@ -30,7 +30,7 @@ export const env = createEnv({
     VITE_JWT_EXPIRATION_MINUTES: z.coerce.number(),
     VITE_GAS_BUFFER_DIVISOR: z.coerce.number(),
     VITE_DOMAIN: z.string().optional(),
-    VITE_ENV: z.enum(['staging', 'production']).default('staging').optional(),
+    VITE_ENV: z.enum(['development', 'staging', 'production']).default('staging').optional(),
```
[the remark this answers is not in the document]
awisniew207: I think this is fine. We actually had a `development` enum previously, but I removed it since I rarely used it. Looks good though, we can merge this

PR #291, opened 2025-08-19
Feat: Separate Vincent dev contract deployment

DashKash54:
# Description

Separate dev contracts where any contracts will be staged and tested before rolling it out to the prod contracts. Also updated the constants to use the dev contracts for testing. Even though this separation is supposed to help us internally move faster anyone using the `getTestClient` will use our dev contracts

## Type of change

- [] Any docs changes have been made

# How Has This Been Tested?

Re-ran Contracts-sdk tests against the dev contracts and it passes.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings


MaximusHaximus (changes requested):
Two things:

getTestClient() is intentionally meant to be for ephemeral, non permanent and not guaranteed data -- so IMO, having it change contract address isn't really a breaking change; we don't guarantee things in the test client are stable/permanent anyway, intentionally
Right now, getTestClient() is only useable as a test target for the registry and the contracts sdk's own tests. We can't use getTestClient() in our ability e2e test suites yet, because the LIT action wrapper code that is bundled uses helper methods in the ability-sdk that internally, in the LIT action, call getClient()  -- so if you use getTestClient() while setting up the delegations for an e2e test that calls an ability, the ability will fail w/ 'delegation doesn't exist for app x' when it, internally, uses getClient() to check the delegation.

We will create a createTestAbilityClient() to handle #2 Soon:tm: -- our ability iterative testing process is pretty painful right now.

DashKash54:
Closing the PR in favor of the simplified [PR](https://github.com/LIT-Protocol/Vincent/pull/305)

PR #295, opened 2025-08-22
Refactor `getAllPermittedAppIdsForPkp` to optimize frontend user dashboard

spacesailor24:
# Description

**This is a breaking change to the contract API. The `getAllPermittedAppIdsForPkp` has been renamed to `getPermittedAppsForPkps`, it's input parameters and return value has changed.**

My understanding of the problem is: a User PKP can have multiple Agent PKPs, and each Agent PKP can have multiple permitted Apps (the contract support it even if we intended there to be 1 Agent PKP per App). When the User loads the App Dashboard, the frontend would ideally make a single RPC request with a list of Agent PKP identifiers (pkp token IDs), and get back the permitted Apps with versions for each Agent PKP, ideally with a flag that signals the frontend on whether or not the permitted App version has been disabled so it can flag that to the User.

To address this, this PR:

- Refactors the existing `getAllPermittedAppIdsForPkp` functions which just returned an array of App IDs permitted by an Agent PKP
    - Function renamed to `getPermittedAppsForPkps`
    - Now accepts an array of PKP token IDs
    - Now returns an array of structs with nested structs instead of an array of App versions.

The new return value is:

```
[
    {
        pkpTokenId: uint256,
        permittedApps: [
            {
                appId: uint40,
                version: uint24,
                versionEnabled: bool
            }
        ]
    }
]
```

One missing property that _may_ save the frontend from making extra RPC calls is the latest version of each permitted App, but I omitted this as I'm not sure we need it.

The rationale behind this refactor:

- We _could_ add new view methods and keep the existing `getAllPermittedAppIdsForPkp` to only accept a single PKP token ID and return an array of just App IDs, but it seems more appropriate to refactor the method to support our needs rather than having two methods that doing basically the same thing, one with just significantly less data.
- We can still fetch a single PKP token ID's permitted Apps by just passing the single token ID if gas cost is a concern (so basically the new function provides the same functionality as before)
- I've opted to not support querying via a PKP ETH address as that would mean we'd need to query the PKP contract to get the token ID since everything is tracked via token ID in the contracts
    - Although, querying the PKP token ID using it's ETH address would be better than the frontend having to do it via separate RPC calls as the Lit nodes would be reading from their local node DB within the single RPC call to `getAllPermittedAppIdsForPkp`
    - However, I'm not sure what info the frontend has available to it, and it seems like our off-chain DB should have the PKP ETH address and token ID stored, so maybe it doesn't have to query the PKP token ID if all it has is the ETH address?

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

- Updated existing assertions in existing test cases

# Checklist:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "packages/libs/contracts-sdk/lib/forge-std"]
+	path = packages/libs/contracts-sdk/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std.git
--- a/packages/libs/contracts-sdk/contracts/VincentDiamond.sol
+++ b/packages/libs/contracts-sdk/contracts/VincentDiamond.sol
@@ -214,13 +214,14 @@ contract VincentDiamond {
     }
 
     function getVincentUserViewFacetSelectors() internal pure returns (bytes4[] memory) {
-        bytes4[] memory selectors = new bytes4[](6);
+        bytes4[] memory selectors = new bytes4[](7);
         selectors[0] = VincentUserViewFacet.getAllRegisteredAgentPkps.selector;
         selectors[1] = VincentUserViewFacet.getPermittedAppVersionForPkp.selector;
         selectors[2] = VincentUserViewFacet.getAllPermittedAppIdsForPkp.selector;
         selectors[3] = VincentUserViewFacet.validateAbilityExecutionAndGetPolicies.selector;
         selectors[4] = VincentUserViewFacet.getAllAbilitiesAndPoliciesForApp.selector;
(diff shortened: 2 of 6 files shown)


.gitmodules
spacesailor24, line 1: @MaximusHaximus I know you wanted to purge the gitsubmodules from the repo, and I think this got removed when that happened. However, I need this in order to be able to use forge to compile the contracts

packages/libs/contracts-sdk/lib/forge-std
spacesailor24, line 1: This seems weird, is this just how git submodules work?

PR #322, opened 2025-09-15
fix(dashboard): properly log errors in sentry, add retry mechanism

awisniew207:
# Description

Adds logging and a retry mechanism for adding payees. Works in sentry:
image

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Locally in the development Sentry project.

# Checklist:

### DIFF
--- a/packages/apps/app-dashboard/src/components/user-dashboard/connect/ConnectPage.tsx
+++ b/packages/apps/app-dashboard/src/components/user-dashboard/connect/ConnectPage.tsx
@@ -18,6 +18,7 @@ import { PKPEthersWallet } from '@lit-protocol/pkp-ethers';
 import { litNodeClient, mintPKPToExistingPKP } from '@/utils/user-dashboard/lit';
 import { useJwtRedirect } from '@/hooks/user-dashboard/connect/useJwtRedirect';
 import { BigNumber } from 'ethers';
+import { addPayee } from '@/utils/user-dashboard/addPayee';
 
 interface ConnectPageProps {
   connectInfoMap: ConnectInfoMap;
@@ -25,7 +26,11 @@ interface ConnectPageProps {
   previouslyPermittedPKP?: IRelayPKP | null;
 }
 
-export function ConnectPage({ connectInfoMap, readAuthInfo, previouslyPermittedPKP }: ConnectPageProps) {
+export function ConnectPage({
+  connectInfoMap,
+  readAuthInfo,
+  previouslyPermittedPKP,
+}: ConnectPageProps) {
   const navigate = useNavigate();
   const [localError, setLocalError] = useState<string | null>(null);
   const [localSuccess, setLocalSuccess] = useState<string | null>(null);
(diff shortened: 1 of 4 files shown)


MaximusHaximus (changes requested):
Added comment in-line

packages/apps/app-dashboard/src/components/user-dashboard/connect/ConnectPage.tsx
```diff
@@ -125,13 +137,54 @@ export function ConnectPage({ connectInfoMap, readAuthInfo, previouslyPermittedP
...
-      await addPermittedActions({
-        wallet: userPkpWallet,
-        agentPKPTokenId: agentPKP.tokenId,
-        abilityIpfsCids: Object.keys(selectedFormData),
-      });
+      const client = getClient({ signer: userPkpWallet });
+
+      try {
+        await addPermittedActions({
+          wallet: userPkpWallet,
```
MaximusHaximus, line 144: This looks to be using the user's PKP wallet to sign -- which uses the pkp sign endpoint that requires sponsorshop -- but the call on L160 tries to add the _agent_ PKP to be sponsored using `addPayee`.

Since the userPKP is the one using the LIT nodes to sign here, **_not_** the agent PKP, don't we need the addPayee() call to add the _userPKP_ to the payment contract if this call fails? I think the only time the agent PKP needs RLI sponsorship is when you're using `withdraw` to submit tx's directly as the agent PKP, right?  Should we be calling `addPayee` for both the agentPKP _and_ the userPKP in parallel here?

packages/apps/app-dashboard/src/components/user-dashboard/connect/ConnectPage.tsx
```diff
@@ -125,31 +137,44 @@ export function ConnectPage({ connectInfoMap, readAuthInfo, previouslyPermittedP
...
+      const client = getClient({ signer: userPkpWallet });
+
+      const attemptPermissions = async () => {
+        await addPermittedActions({
+          wallet: userPkpWallet,
+          agentPKPTokenId: agentPKP.tokenId,
+          abilityIpfsCids: Object.keys(selectedFormData),
+        });
+
         await client.permitApp({
```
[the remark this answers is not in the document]
awisniew207: Updated the PR to match this logic

MaximusHaximus (changes requested):
Just one question about the PKP that we're sponsoring usage for here that is a blocke.

As a non-blocking-but-important aside, this PR made me think about gas usage for actually modifying the contract state (which we do when permitApp is called, modifying permissions, and when actually submitting the signed TXs that `addPermissions`).  How many LIT tokens does the userPKP have to pay for those gas costs? Are we in a situation where a user will run out and the attempt to pkpSign will succeed, but the actual transaction sent to the PKP contracts will fail due to not enough LIT tokens balance? 🤔 Do we ever re-top-up the agent PKP's LIT token balance after the initial funding?

packages/apps/app-dashboard/src/components/user-dashboard/connect/ConnectPage.tsx
awisniew207: The UserPKP is funded initially by the Relayer, and then it's requests will be funded by the Payment Delegation DB. I think the initial funding amount is `0.011`

awisniew207: Resolving just to merge this, but ticketing

MaximusHaximus (approved):
:+1: Let's get this merged!

PR #326, opened 2025-09-17
fix: add swap recipient to signed quote validation

spacesailor24:
# Description

Vulnerability:

Before this PR, the signed Uniswap quote validation logic didn't validate that the recipient of the swap was the User Agent Wallet PKP ETH address, allowing the Delegatee to use an arbitrary address during quote generation, which would swap the Agent Wallet funds from the Agent Wallet address to an arbitrary address

Fix:

Because Uniswap quote generation is happening within a Lit Action, we can simply return the provided recipient for the swap quote as part of the signed quote object returned from the Ability's `prepare` Lit Action that gets provided to the `precheck` and `execute` functions.

Then during the execution of the Ability to execute the swap, we enforce that the recipient of the swap is the Delegator's PKP ETH address

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This is a breaking change because it changes adds a new required field to the input parameters of the `precheck` and `execute` functions

# How Has This Been Tested?

Two new test cases have been added to the swap E2E test:

1. Validates that `precheck` fails when any address other than the Delegator PKP ETH address is provided to the Uniswap quote signer `prepare` Lit Action
2. Validates the same but during the execution of the Ability's `execute` function

# Checklist:

- [x] I created a [release plan](https://nx.dev/recipes/nx-release/file-based-versioning-version-plans) (`nx release plan`) describing my changes and the version bump
- [x] I have added tests that prove my fix is effective or that my feature works


packages/apps/abilities-e2e/test-e2e/swap.spec.ts
```diff
@@ -291,6 +291,97 @@ describe('Uniswap Swap Ability E2E Tests', () => {
     // );
   });
 
+  it('should fail precheck when signed quote recipient does not match delegator PKP ETH address', async () => {
+    // Generate a quote with a different recipient address (malicious attempt)
+    const maliciousRecipient = '0x0000000000000000000000000000000000000002';
+    console.log('Generating malicious Uniswap quote for precheck test...');
```
spacesailor24, line 297:
```suggestion
```

PR #331, opened 2025-09-18
fix(dashboard): atomic saving of authentication info

awisniew207:
# Description

Users ran into an issue when #315 was merged, where their authentication state was broken since we would have part of the authentication information in localstorage (`lit-auth-info`, but not `userPKP`). This would create a race condition with `getValidSessionSigs()`, where we would have authentication information, but we wouldn't be able to generate valid session signatures. This caused the page to then continue, look for session signatures (they don't exist), and then put the users back on the authentication page.

This PR has two fixes:
- Added a conditional to `getValidSessionSigs()` where we only check when we have a `userPKP`. If we don't, it means our stored authentication information is corrupted, and we clear it. This is run immediately after attempting to get valid session signatures, so this serves to ensure that we're not storing stale authentication information if the session signature generation fails.
- It's also made the storing of authentication information atomic. I didn't like how we were partially storing and updating the authentication info, since there was no need and we have to manage partial updates and worry about inconsistent states and future user issues. Now for any AuthMethod, we'll store the information ONLY when we have the UserPKP and the authentication information together. Goodbye race conditions, weird auth states, and partial updates!

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Tested locally using all AuthMethods. 

# Checklist:

### DIFF
--- a/packages/apps/app-dashboard/src/components/user-dashboard/auth/ConnectMethods.tsx
+++ b/packages/apps/app-dashboard/src/components/user-dashboard/auth/ConnectMethods.tsx
@@ -8,7 +8,12 @@ import EthWalletAuth from './EthWalletAuth';
 
 interface ConnectProps {
   authWithWebAuthn: (credentialId: string, userId: string) => Promise<void>;
-  authWithStytch: (sessionJwt: string, userId: string, method: 'email' | 'phone') => Promise<void>;
+  authWithStytch: (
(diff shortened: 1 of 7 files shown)


MaximusHaximus (approved):
:+1: LGTM

glitch003 (approved):
nice refactor of saving auth data

PR #336, opened 2025-09-20
Vincent Wrapped Keys

spacesailor24:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


packages/libs/ability-sdk/src/lib/wrapped-keys/getSolanaKeyPairFromWrappedKey.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // Validate the Vincent prefix
+  const VINCENT_PREFIX = 'vincent_';
+  if (!decryptedPrivateKey.startsWith(VINCENT_PREFIX)) {
+    throw new Error(
+      `Private key was not encrypted with salt; all wrapped keys must be prefixed with '${VINCENT_PREFIX}'`,
+    );
+  }
+
+  // Remove the prefix and create the keypair
```
spacesailor24, line 59:
```suggestion
  const decryptedPrivateKey = await Lit.Actions.decryptAndCombine({
    accessControlConditions,
    ciphertext,
    dataToEncryptHash,
    chain: 'ethereum',
    authSig: null,
  });

  const VINCENT_PREFIX = 'vincent_';
  if (!decryptedPrivateKey.startsWith(VINCENT_PREFIX)) {
    throw new Error(
      `Private key was not encrypted with salt; all wrapped keys must be prefixed with '${VINCENT_PREFIX}'`,
    );
  }

```

packages/libs/wrapped-keys/src/lib/lit-actions/lit-actions-global.d.ts
```diff
@@ -0,0 +1,365 @@
+import { ethers as Ethers } from 'ethers';
+
+declare global {
```
MaximusHaximus, line 3: See other comment on replacing this, which mutates `global` for our entire ts project (all packages, all scopes), with explicit declarations like we did in the starter-kit repo.

This can be a follow-up ticket, as I am sure we've got this global d.ts file in multiple places in the repo by now...

packages/libs/wrapped-keys/src/lib/constants.ts
```diff
@@ -0,0 +1,8 @@
+import type { Network, KeyType } from './types';
+
+export const CHAIN_YELLOWSTONE = 'yellowstone' as const;
+export const VINCENT_PREFIX = 'vincent_' as const;
```
MaximusHaximus, line 4: See other comment -- let's keep this `lit_`

packages/libs/contracts-sdk/src/index.ts
```diff
@@ -42,3 +42,7 @@ export type {
 export { getTestClient, clientFromContract, getClient } from './contractClient';
 
 export { createContract } from './utils';
+
+export { getPkpTokenId } from './utils/pkpInfo';
```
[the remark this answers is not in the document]
spacesailor24: I think this is a misunderstanding, AFAIK, we can't use the `isDelegateePermitted` for the ACCs and by not using `getPkpTokenId`, we force the caller of `getVincentRegistryAccessControlCondition` to provide the token ID instead of being able to use the ETH address

packages/libs/wrapped-keys/src/lib/Lit.d.ts
```diff
@@ -0,0 +1,362 @@
...
+      dataToEncryptHash,
+      authSig,
+      chain,
+    }: {
+      accessControlConditions: unknown[];
+      ciphertext: string;
+      dataToEncryptHash: string;
+      chain: string;
+      authSig: any;
+    }): string;
```
spacesailor24, line 260: @MaximusHaximus I copied this file from the link you shared but the type for the object was `string`, so had to correct it, but it's still wrong in the Starter Kit - I can update there as well, assuming this wasn't intended?

packages/apps/ability-sol-transaction-signer/scripts/deploy-lit-action.js
```diff
@@ -0,0 +1,74 @@
...
+
+(async () => {
+  try {
+    const outputFile = 'lit-action.js';
+
+    const generatedDir = path.join(__dirname, '../src/generated');
+    const filePath = path.join(generatedDir, outputFile);
+    if (!fs.existsSync(filePath)) {
+      throw new Error(
+        `Bundled Lit Action code string not found at ${filePath}. Please run pnpx nx run vincent-ability-evm-transaction-signer:action:build first.`,
```
MaximusHaximus, line 18: :typo: old package name here

packages/apps/ability-sol-transaction-signer/src/lib/lit-action-helpers/signSolanaTransaction.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ethers } from 'ethers';
+
+export function signSolanaTransaction({
+  solanaKeypair,
+  transaction,
+}: {
+  solanaKeypair: Keypair;
+  transaction: Transaction | VersionedTransaction;
+}): string {
+  try {
```
MaximusHaximus, line 11: 🤔 maybe we should have used this pattern in the main wrapped keys SDK instead of that optional argument 'isVersioned...' :(

packages/libs/wrapped-keys/src/lib/api/utils.ts
```diff
@@ -0,0 +1,126 @@
...
+    ).toString();
+  }
+
+  if (!agentPkpTokenId) {
+    throw new Error(
+      'The agent wallet address or PKP token ID is required to create an access control conditions',
+    );
+  }
+
+  const functionAbi = {
```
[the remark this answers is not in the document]
MaximusHaximus: sorry, my comment got mangled -- can we just use the SDK method and make this **always** require the eth address for consistency? We're trying to hide pkp token IDs as much as possible in the Vincent ecosystem and this is re-implementing the logic we have tried to hide inside the vincent-contract-sdk (and bleeding our vincent ABI into this package which is also not great)

spacesailor24:
Removed `lit action prepare.js`, not sure where it came from. Ran the generation scripts, and it didn't come back

packages/libs/wrapped-keys/src/lib/api/utils.ts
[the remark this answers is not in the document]
spacesailor24: We must obtain the PKP token ID for the ACCs, and AFAIK there's no way we can substitute the logic with the contracts SDK since the ACCs are calling a method on the Vincent contract that only takes the PKP token ID

I can remove the ability to pass in the token ID as a parameter, but that would still require the use of the contracts sdk to obtain the token ID using the ETH address

---

In [this commit](https://github.com/LIT-Protocol/Vincent/pull/336/commits/a1151f4cb8015ce9a6b0ffdfa5bd5fff875174d7) I removed the PKP token ID as a param and only accept ETH address, then using the contracts sdk to lookup the PKP token ID for the ACCs. I also exported the ABI from contracts sdk to replace the hardcoded ABI for the `isDelegateePermitted` call in the ACCs

PR #339, opened 2025-09-23
chore(dashboard): permission flow error resiliency 

awisniew207:
# Description

During the approval process, we now save a `pendngAppConnectAgentPkp-<appId>` to know whether there's been a failed permissioning with a PKP.
  
We check localstorage for these states at the beginning of the approval process so we know where to start, reusing the existing PKP in localstorage. Once the user has permitted the app, we remove the state from localstorage.

The existing retry mechanisms for `INSUFFICIENT_FUNDS` and `rate limit exceeded` were kept, but now they'll use the same PKP every time!

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Local dev. I've used the development dashboard to ensure errors at these steps are properly bubbled up to Sentry.

# Checklist:


packages/apps/app-dashboard/src/utils/user-dashboard/approvalStateStorage.ts
```diff
@@ -0,0 +1,58 @@
+import { IRelayPKP } from '@lit-protocol/types';
```
MaximusHaximus, line 1: Please use the [useLocalStorage](https://usehooks-ts.com/react-hook/use-local-storage) hook instead of implementing localstorage integration imperatively here -- we've started re-implementing that hook every time we manage some json state in localstorage :)

MaximusHaximus (changes requested):
I think we're tracking a bit _too much_ state here; I think that we really only need to track a pkp that we minted but couldn't yet assign to the app, because we have no other way of identifying that PKP other than tracking it somewhere.

Whether or not permitted actions have been added for the PKP is identifyable from the chain using standard RPC calls, _and_ adding them is idempotent -- so we don't really need to track whether or not we have.  Let's simplify the state management here to literally store 1 thing -- e.g. `pendngAppConnectAgentPkp-<appId>`, and let everything else detect what it needs to, and idempotently ensure state when it can.

PR #347, opened 2025-09-26
fix: remove sponsorDelegateesPayment endpoint

glitch003:
the sponsorDelagateePayment endpoint is not being called, and does not work.  Instead of having a separate endpoint for this, just do it on app creation, since it's simpler

this PR removes this endpoint and moves it into the app creation endpoint.


glitch003:
merging this b/c it's broken and this might fix it and like it won't break it more

PR #349, opened 2025-09-27
Feat/sol contract whitelist policy

Spacesai1or:
# Description

Adds a new Policy for restricting what contracts can be called by transactions signed by the `@lit-protocol/vincent-ability-sol-transaction-signer` Ability

Users whitelist specific Solana program IDs for each available Solana cluster (network i.e. `devnet`, `testnet`, `mainnet-beta`)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Adds the policy to the existing E2E Solana tx signer Ability tests. Verifies a transaction is permitted to sign calling a whitelisted program on `devnet`, but Ability execution is denied when making a signing request to execute the same program on `mainnet-beta` (which has no whitelisted programs)

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works


packages/apps/abilities-e2e/test-e2e/solana-transaction-signer.spec.ts
```diff
@@ -372,21 +449,24 @@ describe('Solana Transaction Signer Ability E2E Tests', () => {
...
+      solContractWhitelistPolicyMetadata.ipfsCid,
+    );
   });
 
   it('should run precheck and validate transaction deserialization', async () => {
     const client = getSolanaTransactionSignerAbilityClient();
     const precheckResult = await client.precheck(
       {
-        rpcUrl: SOL_RPC_URL,
+        rpcUrl: null,
```
spacesailor24, line 469: I'm having to pass `rpcUrl: null` for both `precheck` and `execute` and make the parameter `.optional().nullable()` in both the Policy and Ability schemas in order to support having an optional parameter for a Policy

`rpcUrl` is something we want as a optional param for Policy/Ability `precheck` because otherwise we'd be using the publically available SOL RPC packaged in the web3.js SDK. However this RPC can be throttled and rate limited, so it needs to be possible for the caller to pass their own RPC URL for `precheck` to function. This isn't an issue with `execute` because the Lit nodes have their own private RPC available

However, when defining:

```ts
const ProgramWhitelistPolicy = createVincentAbilityPolicy({
  abilityParamsSchema,
  bundledVincentPolicy,
  abilityParameterMappings: {
    rpcUrl: 'rpcUrl',
    cluster: 'cluster',
    serializedTransaction: 'serializedTransaction',
  },
});
```

this throws an error if `rpcUrl` is `undefined`. The workaround is to pass `null` but this is bad Devx because you must pass `rpcUrl: null` whenever the Policy is enabled for the Ability

